PR #1, opened 2019-02-12
initial commit

dpopp07:
Opening a PR as my initial commit to get feedback on how I went about separating the core code. The files included are those that I determine to be "core" to the SDK - that is, they would be relevant to any SDK generated with our generator. This included the base service class, the IAM token manager, the request wrapper, and general helper methods.

The proposed name for the NPM package is `ibm-cloud-sdk-core` so that is used in this PR.

I kept all of the style and test configuration for consistency.

The design I chose for usage of this module is to have all functionality available as named sub-modules at the top level of the imported object, as I have seen this design in the community. [This is one example](https://github.com/apollographql/apollo-server#installation).

That's sort of a confusing way of saying it, so here is an example of how it would be used in the Watson Node SDK:
> Package would first be installed with `npm install ibm-cloud-sdk-core`
```js
// assistant/v1.ts
// the headers will now look like...
import * as extend from 'extend';
import { RequestResponse } from 'request';
import { BaseService, getMissingParams } from 'ibm-cloud-sdk-core';
import { getDefaultHeaders } from '../lib/common';

// service code would continue unchanged ...
```

`BaseService` and `getMissingParams` are just two of the sub-modules available (although they are probably the two most used) but that's how they would be imported. The collection of these sub-modules make up the "core functionality" available to dependent SDKs.

To see which modules are exported under which names, see the [`index.ts`](https://github.com/IBM/node-sdk-core/pull/1/files#diff-ed009b6b86b017532ef0489c77de5100) file.


germanattanasio:
@dpopp07 give me a few days to look at this. I imagine the code is good but I want to see if the experience we offer to other SDK owners that may not be familiar with it.

dpopp07:
@germanattanasio That's exactly what I was hoping you would do 👍 
Let me know if you have any questions or ideas and we can discuss

.eslintrc.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  "env": {
```
germanattanasio: why the double quote? 😋

.gitattributes
```diff
@@ -0,0 +1,7 @@
+# https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes
```
germanattanasio: Thanks for adding this file 👍👍👍👍

lib/content-type.ts
```diff
@@ -0,0 +1,64 @@
+import { extname } from 'path';
+import { FileObject } from '../lib/helper';
+
+// This module attempts to identify common content-types based on the filename or header
+// It is not exhaustive, and for best results, you should always manually specify the content-type option.
+// See the complete list of supported content-types at
+// https://console.bluemix.net/docs/services/speech-to-text/input.html#formats
+
+// *some* file types can be identified by the first 3-4 bytes of the file
+const headerContentTypes: { [key: string]: string } = {
```
germanattanasio: maybe you need to think of adding non-watson audio format here

lib/content-type.ts
```diff
@@ -0,0 +1,64 @@
...
+  RIFF: 'audio/wav',
+  OggS: 'audio/ogg',
+  ID3: 'audio/mp3',
+  '\u001aEߣ': 'audio/webm' // String for first four hex's of webm: [1A][45][DF][A3] (https://www.matroska.org/technical/specs/index.html#EBML)
+};
+
+const filenameContentTypes: { [key: string]: string } = {
+  '.mp3': 'audio/mp3',
+  '.wav': 'audio/wav',
+  '.flac': 'audio/flac',
```
germanattanasio: maybe you need to think of adding non-watson audio format here

lib/content-type.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @return {String|undefined}
+ */
+const fromFilename = (file: String | ReadableStream| FileObject |Buffer | File): string => {
+  const ext: string = extname(
+    (typeof file === 'string' && file) || file['name'] || ''
+  );
+  return filenameContentTypes[ext];
+}
+
+export default {
```
germanattanasio: are the functions here tested? if we make this generic we can't expect people testing this by testing their generated service. We have to have custom tests for all the code we are marking as generic

test/resources/weather_data_train.csv
```diff
@@ -0,0 +1 @@
...
Will it rain?,conditions
Will we have much snow?,conditions
Are the winds dangerous?,conditions
What is the expected snowfall today?,conditions
Will it be dry?,conditions
Will it be breezy?,conditions
Will it be humid?,conditions
What is today's expected humidity?,conditions
Will the blizzard hit us?,conditions
Is it drizzling?,conditions
```
germanattanasio: do we need this file?

package.json
```diff
@@ -0,0 +1,87 @@
...
+    "@types/is-stream": "~1.1.0",
+    "@types/node": "~10.3.5",
+    "@types/request": "~2.47.1",
+    "buffer-from": "~1.1.0",
+    "dotenv": "^6.2.0",
+    "extend": "~3.0.2",
+    "file-type": "^7.7.1",
+    "isstream": "~0.1.2",
+    "mime-types": "~2.1.18",
+    "request": "~2.87.0",
```
germanattanasio: This core sdk still uses `request`. Maybe we can update it to use `axious`

.eslintrc.js
dpopp07: Shoot, I will change 😅

lib/content-type.ts
dpopp07: Most of these functions have tests but not all of them. I agree that the core code should be super well-tested. I will work on that

test/resources/weather_data_train.csv
dpopp07: I included all of the resources that were used in the tests. If that one is used for JSON training data, it will get removed but not until the next major version

package.json
dpopp07: I want to get this in before the next major release if possible. That would mean still using `request`. I already have a local branch of the core code with all of my "v4" changes that I will release at the same time as Node SDK v4.

dpopp07:
Thanks for the review. I agree, we should have a robust testing suite. I'll merge this and update the Node SDK, then will start working on filling out the tests here.

package.json
germanattanasio: how much time will you have the old version before replacing it with the new one?

dpopp07: I could have this and the Node SDK updated as early as tomorrow or next week. So as long as it takes us to release v4 of the Node SDK

PR #3, opened 2019-03-18
add docs to readme

dpopp07:
This PR adds documentation to the README for all of the functionality available in this package. This is to help new users get started and get familiar with the exposed modules.

I realized after I wrote it that the documentation could have a nicer, more consistent format so that can be a future change or I can put that in now if anyone feels strongly about it.

I also added the Token Manager class to the top level list of exports, as users certainly may want to use it as a standalone package.

The last fix is just a TypeScript return type that was `string[]` when it should have been `null`.

@padamstx Just requesting you in case you have an opinion on what our near-future users will be seeing.


README.md
```diff
@@ -1,3 +1,86 @@
...
-This project contains the core functionality used by Node SDK's generated by the IBM OpenAPI 3 SDK Generator (openapi-sdkgen).
-Code generated by openapi-sdkgen will depend on the function contained in this project.
+This project contains the core functionality used by Node SDK's generated by the IBM OpenAPI 3 SDK Generator (`openapi-sdkgen`).
+Code generated by `openapi-sdkgen` will depend on the functionality contained in this project.
+
+## Installation
+`npm install ibm-cloud-sdk-core`
+
+## Usage
+This package exports a single object containing a number of modules as top level proerties. This object can be imported all at once or specific
```
padamstx: s/proerties/properties/

Not sure what "all at once or specific" means exactly.

README.md
```diff
@@ -1,3 +1,86 @@
...
+**contentType?**: string
+
+### toLowerKeys
+This function takes an Object and returns the same object with all of the top-level keys converted to lower case. Note: it does not convert nested keys.
+
+### qs
+This module includes one function, `stringify`.
+
+#### qs.stringify
+This functions takes an Object containing query parameters and returns a URI-encoded query String. This function is modified for use with Watson - the query parameter `watson-token` will not be encoded, as the server requires non-encoded tokens for authentication.
```
padamstx: s/functions/function/

README.md
```diff
@@ -1,3 +1,86 @@
...
+
+### contentType
+This module includes two functions, `fromHeader` and `fromFilename`.
+
+#### contentType.fromHeader
+This function takes a Buffer and returns a String. It reads the first bytes of the buffer and attempts to return the mime type associated with the file. Returns `undefined` if no matching mime type is found. Supported types: `audio/flac`, `audio/wav`, `audio/ogg`, `audio/mp3`, `audio/webm`
```
padamstx: where would the user obtain this "Buffer" from exactly?   this detail might help folks to understand how/if to use this function.

dpopp07:
Addressed the desired changes. I also shortened the descriptions to be just summaries rather than trying to fit documentation-style wording in there. I will add a to-do item to add more formal documentation later.

PR #4, opened 2019-03-22
refeactor: make error handling consistent with other sdks

dpopp07:
Make error handling consistent for next major release. First look in `errors[0].message`, then `error`, then `message`. Use the default status text if not found.

Tests included.

### DIFF
--- a/lib/requestwrapper.ts
+++ b/lib/requestwrapper.ts
@@ -67,6 +67,39 @@ function isAuthenticationError(error: any): boolean {
 }
 
 /**
+ * Return true if object has a specified property that is a string
+ * @private
+ * @param {Object} obj - object to look for property in
+ * @param {string} property - name of the property to look for
+ * @returns {boolean} true if property exists and is string
+ */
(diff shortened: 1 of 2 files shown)


lib/requestwrapper.ts
```diff
@@ -66,6 +66,39 @@ function isAuthenticationError(error: any): boolean {
...
+
+  if (Array.isArray(response.errors) && response.errors.length > 0 && hasStringProperty(response.errors[0], 'message')) {
+    message = response.errors[0].message;
+  } else if (hasStringProperty(response, 'error')) {
+    message = response.error;
+  } else if (hasStringProperty(response, 'message')) {
+    message = response.message;
+  }
+
+  return message;
```
mediumTaj: Does this fall back to a generic error if we can't find an error in these places?

dpopp07: Yep! If you look at Line 124, it sets the message to `axiosError.statusText` (which is a generic message for the status code and is always defined) if that function returns `undefined`

PR #10, opened 2019-04-10
fix: parse error message from property errorMessage

dpopp07:
Based on the error format of the IAM service, adding a field to check for when parsing the error - `errorMessage`. 

cc @joyychang


test/unit/requestWrapper.test.js
```diff
@@ -17,6 +17,7 @@ describe('formatError', () => {
         ],
         message: 'Cant find the model.',
         error: 'Model not found.',
+        errorMessage: 'There is just no finding this model.',
```
mediumTaj: 🤣

PR #11, opened 2019-04-10
Improve unit tests

joyychang:


germanattanasio:
@joyychang you need to configure your public github profile to match your git configuration

test/unit/requestWrapper.test.js
```diff
@@ -1,6 +1,279 @@
...
+        formData: '',
+        qs: {},
+        method: 'POST',
+        rejectUnauthorized: true,
+        url:
+          'https://example.ibm.com/v1/environments/environment-id/configurations/configuration-id',
+        headers: {
+          'test-header': 'test-header-value',
+        },
+        responseType: 'jsonp',
```
dpopp07: `jsonp` should be `json`

test/unit/requestWrapper.test.js
```diff
@@ -1,6 +1,279 @@
...
+    sendRequest(parameters, (err, body, res) => {
+      // assert results
+      expect(err).toEqual(expect.anything());
+      expect(body).toBeUndefined();
+      expect(res).toBeUndefined();
+      done();
+    });
+  });
+
+  it('should send a request with overwritten parameters', done => {
```
dpopp07: I think it would be helpful to add a comment somewhere in this test stating that `options` should have the priority over `defaultOptions` for clarity

test/unit/requestWrapper.test.js
```diff
@@ -1,6 +1,279 @@
...
+      expect(axios.mock.calls[0][0].headers['content-type']).toMatch(
+        'multipart/form-data; boundary=--------------------------'
+      );
+      expect(axios.mock.calls[0][0].httpsAgent.options.rejectUnauthorized).toEqual(true);
+      expect(axios.mock.calls[0][0].method).toEqual(parameters.defaultOptions.method);
+      expect(axios.mock.calls[0][0].params).toEqual(parameters.options.qs);
+      expect(axios.mock.calls[0][0].responseType).toEqual('json');
+      expect(res).toEqual('res');
+      expect(axios.mock.calls.length).toBe(1);
+      done();
```
dpopp07: In this test, is there a way we can test that the actual multiform processing is correct? e.g. that the `null_item` is deleted, the `array_item` is stringified, etc? It should send the multiform object in the `data` field

test/unit/requestWrapper.test.js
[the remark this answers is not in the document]
dpopp07: Unless you're trying to test passing in a custom value, which is a good idea. If that's the case, will you change it to `buffer` or `stream` to be more clear? Those are other expected values for `responseType`

test/unit/requestWrapper.test.js
```diff
@@ -1,6 +1,279 @@
...
+        },
+        headers: {
+          'test-header': 'override-header-value',
+          'add-header': 'add-header-value',
+        },
+      },
+    };
+
+    axios.mockImplementation(requestParams => {
+      requestParams.paramsSerializer(requestParams.params);
```
dpopp07: I see what you're doing here, but I'm not sure why it's necessary. Will you add a comment explaining why this line is needed?

test/unit/requestWrapper.test.js
```diff
@@ -1,6 +1,279 @@
...
+    });
+
+    sendRequest(parameters, (err, body, res) => {
+      // assert results
+      expect(axios.mock.calls[0][0].data).toEqual('a=a&b=b');
+      expect(axios.mock.calls[0][0].url).toEqual(
+        'https://example.ibm.com/v1/environments/environment-id/configurations/configuration-id'
+      );
+      expect(axios.mock.calls[0][0].headers).toEqual({
+        'Accept-Encoding': 'gzip',
```
dpopp07: For the last test, will you add a custom value for `Accept-Encoding` and check for that? Something like `compress` is another possible option. I know we have 100% coverage but I don't see a check for that and it would be good for regressions. Let me know if you did test for that and I missed it

dpopp07:
Also, not that this is a big deal, but `test` is a valid header for Angular commit messages so you can use that instead of `chore(test)` e.g. `test: add sendRequest tests`

PR #17, opened 2019-04-25
fix: allow iam client id and secret to be read from constructor

dpopp07:
@padamstx added support for IAM Client ID and Secret credentials in the Token Manager, but I realized that users wouldn't be able to pass these credentials into service constructors without additions to the Base Service. This PR adds those changes w/ tests.

### DIFF
--- a/iam-token-manager/v1.ts
+++ b/iam-token-manager/v1.ts
@@ -17,6 +17,22 @@
 import extend = require('extend');
 import { sendRequest } from '../lib/requestwrapper';
 
+/**
+ * Check for only one of two elements being defined.
+ * Returns true if a is defined and b is undefined,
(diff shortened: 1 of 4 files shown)


padamstx:
I noticed the same thing with Java and was about to get started on the Node-related changes to make it settable by the SDK user :)

joyychang:
We should add a check in token manager to check if secret is provided with client_id

dpopp07:
@joyychang We use the defaults if the user passes in only Client ID or Secret. So, the request would not actually fail but it won't do what the user might expect. I added a warning to let the user know that this is the case.

@padamstx I didn't catch this before but two of the test titles involved "passing only secret via constructor" - I think one was supposed to be using the setter so I changed that. Let me know if that wasn't the intention

PR #18, opened 2019-04-29
fix: carry user options from constructor to axios parameters

dpopp07:
This allows users to set parameters on the `axios` config object using the service constructor.

Before this change, there is no way for the user to have any control over the `axios` configuration. In v3 of the SDK, we allowed parameters to be passed through to the `request` configuration, so this is consistent with that behavior. Maybe this is not something we want to do but I think it is helpful to the user.

Note that this only carries parameters from the service constructor, so it would apply to all requests. We could potentially make these parameters settable on a per-request basis but that would require more changes and is probably less desirable.

### DIFF
--- a/lib/base_service.ts
+++ b/lib/base_service.ts
@@ -15,7 +15,6 @@
  */
 
 import extend = require('extend');
-import semver = require('semver');
 import vcapServices = require('vcap_services');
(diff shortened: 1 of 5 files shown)


dpopp07:
@joyychang Test added

PR #19, opened 2019-05-01
fix: do not read credentials file in webpack override scenario

joyychang:
Related to: https://github.com/watson-developer-cloud/node-sdk/issues/865


germanattanasio:
I think this will work but we need to test it.
There is an example in the `examples` folder. If that works when you load it from a browser then we can merge this.

joyychang:
I need to add unit test still too.

PR #20, opened 2019-05-07
Refactor bluemix to cloud.ibm

mediumTaj:
This pull request refactors bluemix to cloud.ibm. `getCredentialsFromBluemix()` is a private method. Are we safe refactoring this to `getCredentialsFromIbmCloud`?

### DIFF
--- a/lib/base_service.ts
+++ b/lib/base_service.ts
@@ -298,7 +298,7 @@ export class BaseService {
     // but prefer credentials provided programatically
     _options = extend(
       {},
-      this.getCredentialsFromIBMCloud(this.name),
+      this.getCredentialsFromCloud(this.name),
       this.getCredentialsFromEnvironment(process.env, this.name),
       this.getCredentialsFromEnvironment(readCredentialsFile(), this.name),
       options,
@@ -385,7 +385,7 @@ export class BaseService {
    * @private
    * @returns {Credentials}
    */
-  private getCredentialsFromIBMCloud(vcapServicesName: string): Credentials {
+  private getCredentialsFromCloud(vcapServicesName: string): Credentials {
     let credentials: Credentials;
     let temp: any;
     if (this.name === 'visual_recognition') {
--- a/lib/base_service.ts
+++ b/lib/base_service.ts
@@ -294,11 +294,11 @@ export class BaseService {
       _options = extend(_options, options);
       return _options;
     }
-    // Get credentials from environment properties or Bluemix,
+    // Get credentials from environment properties or IBM Cloud,
     // but prefer credentials provided programatically
     _options = extend(
       {},
-      this.getCredentialsFromBluemix(this.name),
+      this.getCredentialsFromIbmCloud(this.name),
       this.getCredentialsFromEnvironment(process.env, this.name),
       this.getCredentialsFromEnvironment(readCredentialsFile(), this.name),
       options,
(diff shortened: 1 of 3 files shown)


joyychang:
You may need to do a search in node-sdk since the base_service is extended, and i'll expect anything that extends it to be able to call `this.functionName` in the parent.

mediumTaj:
OK so the service can still call `getCredentialsFromBluemix()` even though its `private`? No concept of `protected`?

mediumTaj:
It doesn't look like there are any references to `getCredentialsFromBluemix()` in the Watson SDK. Is changing this considered breaking? Can anybody call this method or only classes that extend `base_service`?

joyychang:
I experimented, and the answer is yes.
You would instantiate an instance of the extending class, or base_service itself. then you can do a `instanceName.getCredentialsFromBluemix()`

lib/base_service.ts
```diff
@@ -294,11 +294,11 @@ export class BaseService {
       _options = extend(_options, options);
       return _options;
     }
-    // Get credentials from environment properties or Bluemix,
+    // Get credentials from environment properties or IBM Cloud,
     // but prefer credentials provided programatically
     _options = extend(
       {},
-      this.getCredentialsFromBluemix(this.name),
+      this.getCredentialsFromIbmCloud(this.name),
```
germanattanasio: `IBM` not `Ibm`. I would call it `FromCloud`

mediumTaj:
@germanattanasio I changed the method name to `getCredentialsFromCloud()`.

@joyychang Is it a bad idea to change the method name then? I don't know node - can I keep two methods `getCredentialsFromBluemix()` and `getCredentialsFromCloud()` for compatibility?

lib/base_service.ts
joyychang: do we need some sort of approval for this nomenclature?

joyychang:
In this case you may have to, and add a deprecation message for the Bluemix one. @germanattanasio should this be consider breaking then?

mediumTaj:
I can add a deprecation message for the bluemix operation. I'd like @dpopp07 to look at this before merging as well just to make sure everything looks ok. There isn't a lot of urgency in merging this since we are targeting next regular release for these changes.

mediumTaj:
Hmm its not clear to me how to add the deprecation message. In the repo I see deprecation in the inline documentation
```
/**
 * IBM Watson&trade; Language Translator translates text from one language to another. The service offers multiple domain-specific models that you can customize based on your unique terminology and language. Use Language Translator to take news from across the globe and present it in your language, communicate with your customers in their own language, and more.
 * 
 * @deprecated Language Translator v3 is now available. The v2 Language Translator API will no longer be available after July 31, 2018. To take advantage of the latest service enhancements, migrate to the v3 API. View the [Migrating to Language Translator v3](https://console.bluemix.net/docs/services/language-translator/migrating.html) page for more information.
 */
```

and as a console warning

```
if (!options['silent']) {
    // eslint-disable-next-line no-console
    console.warn(
    'WARNING: Conversation V1 is deprecated and will be removed in the next major release of the SDK. Use Assistant V1 or Assistant V2.' +
        ' Set {silent: true} to disable this message.'
    );
}
```

Which is preferred?

joyychang:
You will need to do both. javascript doesn't have a compiler that parse through annotations to make them have meaning. I don't think typescript is any different.

PR #21, opened 2019-05-07
fix(IAM): renamed UserOptions iam_secret to iam_client_secret to be consistent with other cores

padamstx:
This PR renames `iam_secret` to `iam_client_secret` and `iamSecret` to `iamClientSecret` so that the node sdk core will be similar to other cores wrt IAM-related terminology.


germanattanasio:
this is a breaking change

padamstx:
@germanattanasio and I discussed this PR and determined that the node-sdk currently is locked in on version 0.1.1 of the node core.   This means that we can "safely" (famous last words) make this change without breaking SDK users.

PR #26, opened 2019-05-28
feat: add new token manager for ICP4D

dpopp07:
Changes:
- Introduced a constructor parameter, `authentication_type`, for specifying the authentication pattern. This will be required for using ICP4D
- A new token manager is added, handling ICP4D-style requests
- I created a new generic class, called `JwtTokenManager` for all token-managing needs. This is a base class for the new ICP token manager. I also refactored the IAM token manger to extend the new JWT class too

I still need to write tests but want to start getting feedback on the implementation now, so I don't have to re-write the tests too many times

TODO:
- [x] Write tests
- [x] Add documentation to the README for using `authentication_type`
- [ ] Add a second URL parameter for ICP4D authentication

cc @ehdsouza @mediumTaj @lpatino10 @mamoonraja


auth/jwt-token-manager.ts
```diff
@@ -0,0 +1,169 @@
+/**
+ * Copyright 2015 IBM Corp. All Rights Reserved.
```
mkistler: Need to fix copyright dates.

auth/jwt-token-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /**
+   * Compute and return a Basic Authorization header from a username and password.
+   *
+   * @param {string} username - The username or client id
+   * @param {string} password - The password or client secret
+   * @returns {string}
+   */
+  protected computeBasicAuthHeader(username, password): string {
```
mkistler: This method feels out of place in this file, since the token manager delegates the request processing -- along with any authentication needed -- to its subclasses.

auth/jwt-token-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /**
+   * Request a JWT using an API key.
+   *
+   * @param {Function} cb - The callback that handles the response.
+   * @returns {void}
+   */
+  protected requestToken(cb: Function): void {
+    cb(null, 'token');
```
mkistler: I think the intent is that subclasses override this method, so maybe this method should raise an exception.

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (options.password) {
+      this.password = options.password;
+    }
+    // username and password are required too, unless there's access token
+  }
+
+  /**
+   * Request an ICP token using a basic auth header.
+   *
+   * @param {Function} cb - The callback that handles the response.
```
mkistler: Does Typescript provide a way to document the parameters that will be passed to `cb`?  If there's not a formal method, can we just add this information to the description?

iam-token-manager/v1.ts
```diff
@@ -186,99 +149,4 @@ export class IamTokenManagerV1 {
...
-	let clientSecret = 'bx';
-	
-	// If both the clientId and secret were specified by the user, then use them.
-	if (this.iamClientId && this.iamClientSecret) {
-      clientId = this.iamClientId;
-      clientSecret = this.iamClientSecret;
-	}
-    const encodedCreds = Buffer.from(`${clientId}:${clientSecret}`).toString('base64');
-    return `Basic ${encodedCreds}`;
-  }
```
mkistler: Love all this deleted code!!

auth/jwt-token-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+    const { timeToLive, expireTime } = this;
+
+    if (!timeToLive || !expireTime) {
+      return true;
+    }
+
+    const fractionOfTtl = 0.8;
+    const currentTime = getCurrentTime();
+    const refreshTime = expireTime - (timeToLive * (1.0 - fractionOfTtl));
+    return refreshTime < currentTime;
```
mkistler: Probably shouldn't call this "refreshTime" here, since we've dropped the refresh logic.

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+  private password: string;
+
+  /**
+   * ICP Token Manager Service
+   *
+   * Retreives, stores, and refreshes ICP tokens.
+   *
+   * @param {Object} options
+   * @param {String} options.icpApikey
+   * @param {String} options.icpAccessToken
```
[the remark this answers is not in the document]
dpopp07: Copy/paste error - I will fix that

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @param {String} options.url - url of the icp api to retrieve tokens from
+   * @constructor
+   */
+  constructor(options: Options) {
+    super(options);
+
+    this.tokenName = 'accessToken';
+
+    if (this.url) {
+      this.url = this.url + '/v1/preauth/validateAuth';
```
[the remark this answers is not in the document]
dpopp07: The idea is that when a user is using ICP4D - they pass in their cluster URL. This URL is used to make the standard requests _and_ the authentication request, at a certain endpoint. No reason for them to pass in something like
```
url: 'mycluster.com`,
icpUrl: 'mycluster.com/v1/preauth/validateAuth,
```
when we know what that endpoint is going to be in advance. This is how I understand it, at least.

auth/jwt-token-manager.ts
dpopp07: The `computeBasicAuthHeader` method? It is used by both classes, so I thought it made sense to put it here where it would not be duplicated. It's more of a shared helper method

auth/jwt-token-manager.ts
dpopp07: That's not a bad idea. It needs to be overriden to do anything meaningful but I need to have the method there for it to compile.

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+export type Options = {
+  url: string;
+  accessToken?: string;
+  username?: string;
+  password?: string;
+}
+
+// this interface is a representation of the response
+// object from the ICP service
+export interface IcpTokenData {
```
germanattanasio: `ICP` not `Icp`

auth/icp-token-manager.ts
dpopp07: I don't think there'a formal method but I will check. Either way, yes, I will document it somehow.

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+  sub: string;
+  iss: string;
+  aud: string;
+  uid: string;
+  _messageCode_: string;
+  message: string;
+  accessToken: string;
+}
+
+export class IcpTokenManagerV1 extends JwtTokenManager {
```
germanattanasio: `JWTTokenManager` not `JwtTokenManager`

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+  private username: string;
+  private password: string;
+
+  /**
+   * ICP Token Manager Service
+   *
+   * Retreives, stores, and refreshes ICP tokens.
+   *
+   * @param {Object} options
+   * @param {String} options.icpApikey
```
germanattanasio: `icpApikey` vs `apikey`?

auth/icp-token-manager.ts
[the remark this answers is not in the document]
germanattanasio: v1 will always be in `/v1/preauth/validateAuth` and if we have a new endpoint in `v1` we don't need to request another `url`. That's why base urls are good.

auth/icp-token-manager.ts
germanattanasio: You can call it `callback`

auth/icp-token-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @param {Function} cb - The callback that handles the response.
+   * @returns {void}
+   */
+  protected requestToken(cb: Function): void {
+    const parameters = {
+      options: {
+        url: this.url,
+        method: 'GET',
+        headers: {
+          Authorization:
```
germanattanasio: you need to configure the SSL certificate I think. How are you handling that?

auth/icp-token-manager.ts
dpopp07: Google JS style guide says not to capitalize acronyms - https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

I would prefer to follow this if possible

auth/icp-token-manager.ts
dpopp07: So far, I haven't done anything with that and all the requests have worked. At first, I had `rejectUnauthorized` set to `false` like we do with normal ICP but then I tried it with `true` and it still worked. So, I took it out since it is technically more secure this way. If you think it needs to be in there, I can put it back in.

auth/icp-token-manager.ts
dpopp07: ICP doesn't even use an API key. This was just a mistake on my part

lpatino10:
Should we rename the `IcpTokenManagerV1` class to something like `Icp4dTokenManagerV1`? The name's kind of obnoxious but I think it might be more clear, since ICP and ICP4D are different and use different auth methods. I could see someone getting tripped up by that.

dpopp07:
It is ICP4D specific, so that's definitely not a bad idea. I'll make that change, unless anyone disagrees

auth/jwt-token-manager.ts
dpopp07: I hear you but am not sure what to call it. Going with `timeForNewToken` for now. Let me know if you have other ideas!

auth/icp-token-manager.ts
germanattanasio: sure

auth/icp-token-manager.ts
germanattanasio: I'm worried that you won't always talk to a secure endpoint since the customer can use whatever domain and port they want. We may have to add support for disabling the ssl validation

dpopp07: That's a good point. I can add support for disabling ssl validation but I can also disable it by default if you think we should do that

dpopp07: Okay I just pushed code that carries the `disable_ssl_verification` parameter through to the token managers if the user sets in the service constructor

dpopp07:
Okay last minute (small) changes have been made and all tests have been added. Officially ready for final review

@germanattanasio @mkistler @padamstx

auth/iam-token-manager-v1.ts
```diff
@@ -0,0 +1,158 @@
+/**
+ * Copyright 2015 IBM Corp. All Rights Reserved.
```
mkistler: Need to check copyright date on this file.

auth/jwt-token-manager-v1.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Request a JWT using an API key.
+   *
+   * @param {Function} cb - The callback that handles the response.
+   * @returns {void}
+   */
+  protected requestToken(cb: Function): void {
+    const err = new Error('`requestToken` MUST be overridden by a subclass of JwtTokenManagerV1.');
```
mkistler: Nice!

auth/jwt-token-manager-v1.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Compute and return a Basic Authorization header from a username and password.
+   *
+   * @param {string} username - The username or client id
+   * @param {string} password - The password or client secret
+   * @returns {string}
+   */
+  protected computeBasicAuthHeader(username, password): string {
```
mkistler: I still think this method is out of place in this file.  Maybe it belongs in it's own file: `auth/basic-v1.ts`?

lib/base_service.ts
```diff
@@ -56,23 +59,30 @@ export interface Credentials {
...
+      obj.icp_access_token)
   );
 }
 
 function isForICP(cred: string): boolean {
   return cred && cred.startsWith('icp-');
 }
 
+function isForICP4D(obj: any): boolean {
+  return obj && (obj.authentication_type === 'icp4d' || obj.icp_access_token);
```
mkistler: This isn't consistent with the README.md where it says:
> In order to use ICP4D authentication, this option [`authentication_type`] **must** be passed in.

auth/iam-token-manager-v1.ts
dpopp07: Yes, that's certainly wrong. The IAM token manager was originally written in 2018 but this specific file in this repo was added in 2019. So - should it be `2018, 2019` or `2019`?

auth/jwt-token-manager-v1.ts
dpopp07: That's a good idea. This logic is used in the base class too, so putting it in a utility type file and sharing it makes sense.

lib/base_service.ts
dpopp07: Oh, good point. I decided after writing the docs that `icp_access_token` is sufficient for determining if `icp4d` is the authentication type. I'll update the docs to reflect that.

Also, I wonder if that variable should be called `icp4d_access_token` for consistency...

dpopp07:
We're not using _refresh tokens_ anymore. We are requesting a new token when the token is expired and I might have referred to that as a "refresh" but I agree that may be misleading. I'll update that

lib/base_service.ts
```diff
@@ -128,6 +137,8 @@ export class BaseService {
    * @param {string} [options.iam_url] - url for iam service api, needed for services in staging
    * @param {string} [options.iam_client_id] - client id (username) for request to iam service
    * @param {string} [options.iam_client_secret] - secret (password) for request to iam service
+   * @param {string} [options.icp4d_access_token] - icp access token provided and managed by user
```
germanattanasio: icp or icp4d?

germanattanasio: if the parameter includes 4d then you should say `ICP for Data`

germanattanasio:
@dpopp07 can you add an icon to `ibm-devx-automation` 😛 ?

joyychang:
no

dpopp07:
Haha I would if I could 🤷‍♂

dpopp07:
@germanattanasio Oh wait, it looks like it does have an icon!

PR #27, opened 2019-05-29
Axios

joyychang:


lib/requestwrapper.ts
```diff
@@ -20,23 +20,276 @@ import FormData = require('form-data');
...
 const isBrowser = typeof window === 'object';
 const globalTransactionId = 'x-global-transaction-id';
 
-// axios sets the default Content-Type for `post`, `put`, and `patch` operations
-// to 'application/x-www-form-urlencoded'. This causes problems, so overriding the
-// defaults here
-['post', 'put', 'patch'].forEach(method => {
-  axios.defaults.headers[method]['Content-Type'] = 'application/json';
-});
+const allowedAxiosConfig = ['transformRequest', 'transformResponse', 'paramsSerializer', 'paramsSerializer', 'timeout', 'withCredentials', 'adapter', 'responseType', 'responseEncoding', 'xsrfCookieName', 'xsrfHeaderName', 'onUploadProgress', 'onDownloadProgress', 'maxContentLength', 'validateStatus', 'maxRedirects', 'socketPath', 'httpAgent', 'httpsAgent', 'proxy', 'cancelToken'];
```
dpopp07: So I don't _love_ the idea of maintaining a list of axios request params in this class. If we don't want certain things to make it in the request, I would rather blacklist them than whitelist all of these.

That said, if you think this is best I'm fine to leave as is. Maybe throw a comment in there saying why you're doing it that way

lib/requestwrapper.ts
```diff
@@ -20,23 +20,276 @@ import FormData = require('form-data');
...
+    extend(true, axiosConfig, allowedAxiosConfig.reduce((a,b)=> (a[b]=axiosOptions[b], a),{}));
+
+    this.axiosInstance = axios.create(axiosConfig);
+
+    // set debug interceptors
+    if(process.env.NODE_DEBUG === 'axios') {
+      this.axiosInstance.interceptors.request.use(config => {
+        console.debug('Request:');
+        try {
+          console.debug(JSON.stringify(config, null, 4));
```
dpopp07: Will you make these 2 space indents, rather than 4? To be consistent with our style

lib/requestwrapper.ts
```diff
@@ -20,23 +20,276 @@ import FormData = require('form-data');
...
+          'Content-Type':'application/json'
+        },
+        patch: {
+          'Content-Type':'application/json'
+        },
+      }
+    };
+
+    // merge valid Axios Config into default.
+    extend(true, axiosConfig, allowedAxiosConfig.reduce((a,b)=> (a[b]=axiosOptions[b], a),{}));
```
dpopp07: Would you alter this a little for readability? Either name `a` and `b` differently like `opts` and `key`, or just add a comment to describe what it's doing

iam-token-manager/v1.ts
```diff
@@ -92,6 +93,8 @@ export class IamTokenManagerV1 {
       // tslint:disable-next-line
       console.log(CLIENT_ID_SECRET_WARNING);
     }
+
+    this.requestWrapperInstance = new RequestWrapper({});
```
dpopp07: What are your thoughts on altering the constructor so that if no params are passed in, it defaults to `{}`? I hate to make the user pass in an empty object if they aren't providing any parameters

test/unit/requestWrapper.test.js
```diff
@@ -273,65 +272,64 @@ describe('sendRequest', () => {
...
-
-    sendRequest(parameters, (err, body, res) => {
-      // assert results
-      expect(axios.mock.calls[0][0].otherParam).toEqual(500);
-      expect(res).toEqual('res');
-      expect(axios.mock.calls.length).toBe(1);
-      done();
-    });
-  });
+  // Need to rewrite this to test instantiation with userOptions
```
dpopp07: What needs to be done for this? Will you open an issue in the SDK planning board for this so we don't lose it?

lib/requestwrapper.ts
```diff
@@ -20,23 +20,276 @@ import FormData = require('form-data');
 import https = require('https');
 import querystring = require('querystring');
 import { PassThrough as readableStream } from 'stream';
+import {HeaderOptions} from "./base_service";
```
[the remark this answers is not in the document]
joyychang: removed unnecessary extra line.

PR #43, opened 2019-09-15
Fix handling of array form parameters

mkistler:
This PR fixes the handling of array form parameters.

In a `multipart/form-data` request body, an array property should be serialized as multiple form data parts, each with the name of the array property and one of the elements of the array.

I've also added documentation for the primary exported method, `createRequest`, and refactored the processing of file properties to make it simpler/clearer.

### DIFF
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -43,12 +43,21 @@ export interface FileStream extends NodeJS.ReadableStream {
 }
 
 // custom type guards
-function isFileObject(obj: any): obj is FileObject {
+export function isFileObject(obj: any): obj is FileObject {
   return Boolean(obj && obj.value);
 }
 
 function isFileStream(obj: any): obj is FileStream {
   return obj && isReadable(obj) && obj.path;
+}
+
+export function isFileParamAttributes(obj: any): obj is FileParamAttributes {
+  return obj && obj.data &&
+   (
+     isReadable(obj.data) ||
+     Buffer.isBuffer(obj.data) ||
+     isFileObject(obj.data)
+   );
 }
 
 export function isFileParam(obj: any): boolean {
@@ -58,7 +67,7 @@ export function isFileParam(obj: any): boolean {
       isReadable(obj) ||
       Buffer.isBuffer(obj) ||
       isFileObject(obj) ||
-      (obj.data && isFileParam(obj.data))
+      isFileParamAttributes(obj)
     )
   );
 }
@@ -164,6 +173,7 @@ export function getFormat(
  * this function builds a `form-data` object for each file parameter
  * @param {FileParamAttributes} fileParams - the file parameter attributes
  * @param {NodeJS.ReadableStream|Buffer|FileObject} fileParams.data - the data content of the file
+ * @param (string) fileParams.filename - the filename of the file
  * @param {string} fileParams.contentType - the content type of the file
  * @returns {FileObject}
  */
--- a/lib/requestwrapper.ts
+++ b/lib/requestwrapper.ts
@@ -20,7 +20,7 @@ import FormData = require('form-data');
 import https = require('https');
 import querystring = require('querystring');
 import { PassThrough as readableStream } from 'stream';
-import { buildRequestFileObject, getMissingParams, isEmptyObject, isFileParam } from './helper';
+import { buildRequestFileObject, getMissingParams, isEmptyObject, isFileObject, isFileParam, isFileParamAttributes } from './helper';
 
 const isBrowser = typeof window === 'object';
 const globalTransactionId = 'x-global-transaction-id';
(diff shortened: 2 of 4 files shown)


mkistler:
I modified the Watson Node SDK to use this version of the core and ran the full test suite, and that turned up one issue that needs further investigation.  A STT test failed when calling `addCorpus`.  The error response is:
```
Bad Request: Invalid corpus content type. Please use 'multipart/form-data' and 'corpus_file' key to upload the corpus file.
```

mkistler:
I have investigated the failure in the STT `addCorpus` test and it looks like there is a corner case that we have to work out.  The failing STT test calls `addCorpus` as follows:

```
    speech_to_text.addCorpus(
      {
        customization_id: customization_id,
        corpus_name: 'test_corpus_2',
        corpus_file: fs
          .readFileSync(path.join(__dirname, '../resources/speech_to_text/corpus-short-2.txt'))
          .toString(),
        allow_overwrite: true,
      },
      done
    );
```

So the call is passing a plain string as the value for `corpus_file`.  This parameter is defined in the SDK as:
```
 * @param {NodeJS.ReadableStream|FileObject|Buffer} params.corpus_file 
```

So a `string` is technically not a valid input.  But it worked before, so could we make it work now?

It seems that the way it worked before was that the STT SDK constructed a `FileParamAttributes` object and passed the string in the `data` property.  That also isn't valid according to the node-sdk-core, which defines `FileParamAttributes` as:
```
export interface FileParamAttributes {
  data: NodeJS.ReadableStream | Buffer | FileObject;
  contentType: string;
  filename: string;
}
```

But it worked before ... so could we extend the definition of `FileParamAttributes` to allow a `string` to be passed in the `data` property?  Well, if we did, we would break this test in "isFileParam.test.js":
```
  it('should not think object is file just because of `data` property', () => {
    const nonFileObject = { data: 'should not be interpreted as file' };
    expect(isFileParam(nonFileObject)).toBe(false);
  });
```

So ... what to do?

lib/requestwrapper.ts
```diff
@@ -127,48 +127,30 @@ export class RequestWrapper {
...
-      Object.keys(formData).forEach(key => {
-        // handle files differently to maintain options
-        if (formData[key].value) {
-          multipartForm.append(key, formData[key].value, formData[key].options);
-        } else {
-          multipartForm.append(key, formData[key]);
-        }
+        const values = Array.isArray(formData[key]) ? formData[key] : [formData[key]];
+        // Skip keys with undefined/null values or empty object value
+        values.filter(v => v != null && !isEmptyObject(v)).forEach(value => {
```
dpopp07: I think this will miss values of explicit type `undefined`. The method `isEmptyObject` is pretty strict about returning `true` iff the value passed in is the empty object `{}`. 

Something a little looser would catch more values:
```suggestion
        values.filter(v => v && !isEmptyObject(v)).forEach(value => {
```

Or we can just be more explicit:

`values.filter(v => v != null && v != undefined && !isEmptyObject(v)).forEach(value => {`

but I don't know if that's necessary.

lib/helper.ts
```diff
@@ -43,22 +43,31 @@ export interface FileStream extends NodeJS.ReadableStream {
...
+export function isFileObject(obj: any): obj is FileObject {
   return Boolean(obj && obj.value);
 }
 
 function isFileStream(obj: any): obj is FileStream {
   return obj && isReadable(obj) && obj.path;
 }
 
+export function isFileParamAttributes(obj: any): obj is FileParamAttributes {
+  return obj && obj.data &&
```
dpopp07: I don't think this is a big deal but this should maybe be wrapped in a `Boolean()` cast to ensure the result is either `true` or `false`. The way is stands, it could end up returning `undefined` if `obj` or `obj.data` is undefined. That is fine, because it still evaluates to falsy, but it is probably more clear to force the boolean type.

lib/helper.ts
```diff
@@ -43,22 +43,31 @@ export interface FileStream extends NodeJS.ReadableStream {
...
-function isFileObject(obj: any): obj is FileObject {
+export function isFileObject(obj: any): obj is FileObject {
   return Boolean(obj && obj.value);
 }
 
 function isFileStream(obj: any): obj is FileStream {
   return obj && isReadable(obj) && obj.path;
 }
 
+export function isFileParamAttributes(obj: any): obj is FileParamAttributes {
```
dpopp07:
```ts
obj is FileParamAttributes
```

Cool, I've never seen this! Does this imply a Boolean return value? Did you find any docs/posts about this and if so, could you link me? Looks interesting

lib/requestwrapper.ts
mkistler: The filter as coded will eliminate "undefined" values.
```
[mkistler@mkistlers-mbp] ~>node
> foo = [ undefined ]
[ undefined ]
> foo.filter(v => v != null)
[]
```

Is there another case you are concerned about?

dpopp07: Ah, my mistake. I was thinking about the results when using the more strict `!==` rather than `!=`. This looks good as is 👍

PR #47, opened 2019-09-19
fix: Move check for serviceUrl to createRequest

mkistler:
This PR moves the check for a non-empty serviceUrl to the createRequest method, so that sendRequest can be used from the authenticators which just pass `url`.

In addition, createRequest will take serviceUrl from the service baseOptions if none was provided in  parameters.defaultOptions.

Tests were updated accordingly.


mkistler:
On further consideration, I decided to drop the logic that grabs the serviceUrl from baseOptions if one is not specified. That was a half solution that isn't needed here and should be addressed fully in a separate PR (in the future).

PR #48, opened 2019-09-19
Fix backward compatibility for FileObjects 

mkistler:
The PR fixes the backward incompatibility for FileObjects passed as files in formData.

I removed support for FileObjects in a previous PR because I thought it was not needed or used, but  subsequently discovered that some users do use it.  So I've reimplemented support for FileObjects and restored all the tests I had deleted in the previous PR.

### DIFF
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -33,8 +33,8 @@ export interface FileOptions {
 
 export interface FileWithMetadata {
   data: NodeJS.ReadableStream | Buffer;
-  contentType: string;
   filename: string;
+  contentType: string;
 }
 
 export interface FileStream extends NodeJS.ReadableStream {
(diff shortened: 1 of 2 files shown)


lib/helper.ts
```diff
@@ -162,35 +166,47 @@ export function getFormat(
...
   // build contentType
-  let contentType: string = 'application/octet-stream';
-  if (fileParam.contentType) {
-    // prioritize user-given content_type
-    contentType = fileParam.contentType;
-  } else {
-    // else utilize file-type package
-    contentType = getContentType(fileParam.data) || contentType;
+  if (!fileObj.options.contentType && isFileData(fileObj.value)) {
+
```
dpopp07: This is picky I know, but can we remove this newline? The TS linter won't complain about but technically it violates the style guide

PR #49, opened 2019-09-22
minor refactoring for the authentication changes

dpopp07:
* refactor validate method to be used across authenticators/token mangers
* add types for token manager properties in authenticators
* add return type for `getAuthenticatorFromEnvironment`
* return `Error` object in `checkCredentials` rather than a `string`
* add setters for `disableSslVerification` and `headers` in token managers so we can use stricter typing
* Make `authenticator` optional in the base service options

This last point needed to be done to compile the Watson SDKs - TypeScript does not allow you to declare an optional parameter on an interface that extends another interface where the same parameter is required. We aren't _exactly_ extending the interface (although I think we should) but we use the "extended" interface in a place where the base interface is expected and TS complains. This resolves the issue. We still validate for the authenticator, just can't type it that way.

### DIFF
--- a/auth/authenticators/authenticator.ts
+++ b/auth/authenticators/authenticator.ts
@@ -15,8 +15,6 @@
  */
 
 import { OutgoingHttpHeaders } from 'http';
-import { getMissingParams } from '../../lib/helper';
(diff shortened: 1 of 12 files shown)


dpopp07:
It does and actually needs to PR in order to compile. I'll create an RC to use in the compatibility chart

PR #50, opened 2019-09-23
feat: export `UserOptions` interface from the BaseService

dpopp07:
Export this type to be used in the generated subclasses.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -18,7 +18,7 @@
  * @module ibm-cloud-sdk-core
  */
 
-export { BaseService } from './lib/base_service';
+export { BaseService, UserOptions } from './lib/base_service';
 export * from './auth';
 export * from './lib/helper';
 export { default as qs } from './lib/querystring';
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -18,7 +18,7 @@
  * @module ibm-cloud-sdk-core
  */
 
-export { BaseService } from './lib/base_service';
+export { BaseService, UserOptions } from './lib/base_service';
```
mkistler: If you are going to export `UserOptions`, should you not also export `HeaderOptions`?

dpopp07:
That's a good point. I actually don't see `HeaderOptions` defined in the code anywhere. Only in the Typedoc comments, where it definitely needs to be changed. We can also move the documentation of the learning opt out header to Watson, although we would need a way to represent that in the API definition.

I'll go ahead and remove that from the core for now. I'll also add comments for all of the interface fields 👍

dpopp07:
@mkistler Comments addressed. I also re-ordered the properties of the interface

PR #51, opened 2019-09-25
refactor(types): make user options extend axios request options

dpopp07:
* this removes the need for allowing arbitrary additional properties without adding too much axios-specific code
* also, adds types to a few other local variables

It occurred to me this morning that, since `axios` defines their own types, we could use their request options as strict parameters instead of allowing additional properties on the base service constructor options. I think this makes for a much better TypeScript experience, since the additional properties makes the existing types worth little more than documentation.

I also went ahead and added some other types to a few local variables in the request wrapper class. Minor changes but the more types we define the more benefit we get out of using TS.


lib/base_service.ts
```diff
@@ -23,7 +23,7 @@ import { AuthenticatorInterface, checkCredentials, readExternalSources } from '.
 import { stripTrailingSlash } from './helper';
 import { RequestWrapper } from './requestwrapper';
 
-export interface UserOptions {
+export interface UserOptions extends AxiosRequestConfig {
```
mkistler: Given that we export `UserOptions` from this package, doesn't making it extend `AxiosRequestConfig` expose the underlying transport mechanism of the core?  If so, I think that's not something we want to do.

dpopp07:
After some discussion with @mkistler, we are going to close this and re-evaluate the possibility of writing our own interface on top of whatever request library we may be using

PR #52, opened 2019-09-27
chore: replace @types/is-stream with correct @types/isstream

MasterOdin:
This changes the installed `@types` file to use the right typing package for the isstream module that is depended upon concretely.

##### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1574,13 +1574,10 @@
         "@types/node": "*"
       }
     },
-    "@types/is-stream": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/@types/is-stream/-/is-stream-1.1.0.tgz",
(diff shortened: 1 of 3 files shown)


MasterOdin:
It seems like `file-type` is also missing from the `types` array in the tsconfig.json, though I've never used it so I'm not sure if that was on purpose or not.

dpopp07:
I didn't do that on purpose, so if you think it would be useful feel free to add it

MasterOdin:
Would it also make sense to just remove the key and let `tsc` automatically find the necessary types it needs automatically from the `node_modules/@types/` directory?

dpopp07:
If TypeScript does that, yeah that sounds good to me

MasterOdin:
Sounds good, I've commented out that bit of code.

PR #54, opened 2019-09-27
Handle ssl errors

dpopp07:
Adds a specific handler for ICP SSL errors. Also fixes the dreaded "Body is HTTP blah blah" error with a more descriptive one.

### DIFF
--- a/lib/requestwrapper.ts
+++ b/lib/requestwrapper.ts
@@ -258,7 +258,7 @@ export class RequestWrapper {
       } catch (e) {
         // ignore the error, use the object, and tack on a warning
         errorBody = axiosError.data;
-        errorBody.warning = 'body contains circular reference';
+        errorBody.warning = 'Body contains circular reference';
       }
 
       error.body = errorBody;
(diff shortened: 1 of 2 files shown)


mediumTaj:
dustin was this meant to go to `master`?

dpopp07:
Nope, updated

lib/requestwrapper.ts
```diff
@@ -270,14 +270,20 @@ export class RequestWrapper {
...
-      error.body = axiosError.request;
-
+      error.message = axiosError.message;
+      error.statusText = axiosError.code;
+      error.body = 'Response not received - no connection was made to the service.';
+
+      // when a request to a private cloud instance has an ssl problem, it never connects and follows this branch of the error handling
+      if (isSelfSignedCertificateError(axiosError)) {
+        error.message = `If you're trying to call a service on ICP or Cloud Pak for Data, you ` +
+          `may not have a valid SSL certificate. If you need to access the service without setting that up, try using ` +
```
germanattanasio: What about?

> The service you are trying to call doesn't use valid certificates. You can disable the SSL certificate validation using the `disableSslVerification` option.

dpopp07: I'm fine with that. I just copied the message that @lpatino10 has in Java for consistency. Would you rather this change?

SirSpidey:
We usually state the error and the possible response. How about this?

The connection failed because the SSL certificate is not valid. To use a self-signed certificate, set the `disableSslVerification` parameter in`HttpConfigOptions`.

dpopp07:
I'm fine with any of these. I just think that currently, all of the other SDKs are using the message in the PR, so do we want to change all of them? @mediumTaj

mediumTaj:
Yes once we settle i will create an issue

PR #55, opened 2019-10-01
feat: refactor core to use Promises instead of callbacks

dpopp07:
Refactors the entire library to use Promises instead of callbacks. This is a pretty big change but since the SDK is really the only "user", I don't think it will cause any problems.

The generated SDK code should be able to change slightly to achieve no difference in current behavior. The current plan is to start returning Promises first in the generated Watson SDK, with callbacks being an accommodation. We also want to start generating new services using only Promises. This PR enables that route, as it moves all callback handling out of the core and into the generated SDK.

If this is too big of a change to make too late in the game, that's fine, but this work is necessary to move towards a Promise-only approach and doing it this way gives us more flexibility in the generated SDKs to use either approach.

If this is agreeable, I will open a corresponding PR in the generator.

FYI: This PR makes heavy use of the fact that every `then` and `catch` block returns a Promise itself and that "returning" or "throwing" inside of these blocks, will resolve and reject in a later chained Promise, respectively. Where we were previously passing one callback around to multiple functions, we are now bubbling up results through chained Promises. Using the other method, we used to lose some error information that is now becoming visible. This required some small changes in the unit tests. On that point, I think the code is improved overall.

### DIFF
--- a/auth/authenticators/authenticator.ts
+++ b/auth/authenticators/authenticator.ts
@@ -15,7 +15,7 @@
  */
 
 import { OutgoingHttpHeaders } from 'http';
-import { AuthenticateCallback, AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
+import { AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
 
 export class Authenticator implements AuthenticatorInterface {
   /**
@@ -31,7 +31,8 @@ export class Authenticator implements AuthenticatorInterface {
     }
   }
 
-  public authenticate(options: AuthenticateOptions, callback: AuthenticateCallback): void {
-    throw new Error('Should be implemented by subclass!');
+  public authenticate(options: AuthenticateOptions): Promise<void | Error> {
+    const error = new Error('Should be implemented by subclass!');
+    return Promise.reject(error);
   }
 }
--- a/auth/authenticators/basic-authenticator.ts
+++ b/auth/authenticators/basic-authenticator.ts
@@ -17,7 +17,7 @@
 import extend = require('extend');
 import { computeBasicAuthHeader, validateInput } from '../utils';
 import { Authenticator } from './authenticator';
-import { AuthenticateCallback, AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
+import { AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
 
 export type Options = {
   username?: string;
@@ -48,11 +48,13 @@ export class BasicAuthenticator extends Authenticator implements AuthenticatorIn
     this.password = options.password;
   }
 
-  public authenticate(options: AuthenticateOptions, callback: AuthenticateCallback): void {
-    const authHeaderString = computeBasicAuthHeader(this.username, this.password);
-    const authHeader = { Authorization: authHeaderString }
+  public authenticate(options: AuthenticateOptions): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const authHeaderString = computeBasicAuthHeader(this.username, this.password);
+      const authHeader = { Authorization: authHeaderString }
 
-    options.headers = extend(true, {}, options.headers, authHeader);
-    callback(null);
+      options.headers = extend(true, {}, options.headers, authHeader);
+      resolve();
+    });
   }
 }
--- a/auth/authenticators/token-request-based-authenticator.ts
+++ b/auth/authenticators/token-request-based-authenticator.ts
@@ -18,7 +18,7 @@ import extend = require('extend');
 import { OutgoingHttpHeaders } from 'http';
 import { JwtTokenManager } from '../token-managers';
 import { Authenticator } from './authenticator';
-import { AuthenticateCallback, AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
+import { AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
 
 export type BaseOptions = {
   headers?: OutgoingHttpHeaders;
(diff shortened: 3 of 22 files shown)


auth/authenticators/basic-authenticator.ts
```diff
@@ -48,11 +48,11 @@ export class BasicAuthenticator extends Authenticator implements AuthenticatorIn
     this.password = options.password;
   }
 
-  public authenticate(options: AuthenticateOptions, callback: AuthenticateCallback): void {
+  public authenticate(options: AuthenticateOptions): Promise<void> {
     const authHeaderString = computeBasicAuthHeader(this.username, this.password);
```
germanattanasio: I don't know about TypeScript but with JS, you need to always wrap the code within a promise in order to catch errors.

```ts
public authenticate(options: AuthenticateOptions): Promise<void> { 
  return new Promise((resolve, reject) => {
     const authHeaderString = computeBasicAuthHeader(this.username, this.password); 
     const authHeader = { Authorization: authHeaderString } 
     options.headers = extend(true, {}, options.headers, authHeader);
     resolve();
 });
}
```

auth/authenticators/token-request-based-authenticator.ts
```diff
@@ -84,15 +84,14 @@ export class TokenRequestBasedAuthenticator extends Authenticator implements Aut
...
+    return this.tokenManager.getToken()
+      .then(token => {
         const authHeader = { Authorization: `Bearer ${token}` };
         options.headers = extend(true, {}, options.headers, authHeader);
-        callback(null);
-      }
-    });
+        return;
+      }, err => {
+        throw(err);
```
germanattanasio: why the `throw`? isn't the caller supposed to handle the `catch()`?

auth/authenticators/authenticator.ts
```diff
@@ -31,7 +31,8 @@ export class Authenticator implements AuthenticatorInterface {
     }
   }
 
-  public authenticate(options: AuthenticateOptions, callback: AuthenticateCallback): void {
-    throw new Error('Should be implemented by subclass!');
+  public authenticate(options: AuthenticateOptions): Promise<void | Error> {
```
germanattanasio:
```ts
public authenticate(options: AuthenticateOptions): Promise<void | Error>
```
or
```ts
public authenticate(options: AuthenticateOptions): Promise<void>
```

auth/token-managers/jwt-token-manager.ts
```diff
@@ -70,33 +70,30 @@ export class JwtTokenManager {
...
    * 1. If user provides their own managed access token, assume it is valid and send it
    * 2. a) If this class is managing tokens and does not yet have one, make a request for one
    *    b) If this class is managing tokens and the token has expired, request a new one
    * 3. If this class is managing tokens and has a valid token stored, send it
    *
-   * @param {Function} cb - callback function that the token will be passed to
    */
-  public getToken(cb: Function) {
+  public getToken(): Promise<any> {
     if (!this.tokenInfo[this.tokenName] || this.isTokenExpired()) {
```
germanattanasio: I think this should be wrap in a promise like my example above

auth/authenticators/token-request-based-authenticator.ts
dpopp07: All Promise handlers (`then`, `catch`, `finally`) return another Promise. If you return a value inside, it is the same as "resolving" the Promise for the next handler - if you throw an error, it is the same as "rejecting". So as long as the caller anticipates a rejection, this should be fine.

auth/authenticators/basic-authenticator.ts
dpopp07: Promises work the same in TS as they do JS. That is true, I just wasn't anticipating any errors here and wasn't handling them when I was using callbacks. But I am good with this approach here to be safe.

auth/authenticators/authenticator.ts
dpopp07: This needs to hold the superset of the different types for the authenticators

auth/token-managers/jwt-token-manager.ts
dpopp07: Since the `requestToken` method also returns a Promise, I think it is cleaner to chain the Promises together than to create a new Promise altogether.

dpopp07:
In some scenarios I would think that is better but when returning a series of Promises in a row (like `authenticate` -> `getToken` -> `requestToken` -> `sendRequest`) I think it makes more sense to use Promise chaining and, from each method, return the Promise returned from the method before it, with handlers to do the method-specific processing. Let me know what you think

auth/token-managers/jwt-token-manager.ts
germanattanasio: What happens if `this.tokenInfo` is undefined?

auth/authenticators/basic-authenticator.ts
germanattanasio: I'm just trying to anticipate future changes to those methods that today don't return errors

auth/authenticators/token-request-based-authenticator.ts
germanattanasio: I was asking why are you actually doing `,err -> { throw(err);}` I think that if you don't add it then it will be the same

dpopp07: Oh, I see what you're saying. You are correct about that, I will update the code

auth/token-managers/jwt-token-manager.ts
dpopp07: It would throw an error that would not be rejected in the Promise, so the users rejection handling wouldn't catch it. Which would be bad. But we explicitly define the `tokenInfo` object in the constructor: `this.tokenInfo = {};`.

I understand what you're saying, I just don't know how much we want to account for a scenario like this that shouldn't ever happen. We were not accounting for any errors happening here before either. The behavior is staying the same. That said, if you think we need to be able to catch errors there, I will change the code

PR #59, opened 2019-10-03
feat: allow node-sdk-core 0.x to work with new ibm-crendentials.env

MasterOdin:
From my understand and talking with @dpopp07, with the release of the v5 API SDK, the `ibm-credentials.env` file is going to change format from:
```
TEXT_TO_SPEECH_IAM_APIKEY=my_api_key
TEXT_TO_SPEECH_URL=https://stream.watsonplatform.net/text-to-speech/api
```

to be:
```
TEXT_TO_SPEECH_APIKEY=my_api_key
TEXT_TO_SPEECH_URL=https://stream.watsonplatform.net/text-to-speech/api
```

This PR allows the v4 API to consume this new format to make things slightly less painful for those of us who cannot immediately upgrade to v5, but would like to continue downloading these env files when setting up new services in the meantime.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/base_service.ts
+++ b/lib/base_service.ts
@@ -449,7 +449,7 @@ export class BaseService {
       return this.getCredentialsFromEnvironment(envObj, 'visual_recognition');
     }
     // Case handling for assistant - should look for assistant env variables before conversation
-    if (name === 'conversation' && (envObj[`ASSISTANT_USERNAME`] ||  envObj[`ASSISTANT_IAM_APIKEY`])) {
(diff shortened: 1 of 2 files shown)


MasterOdin:
The original usage of the `apiKey` variable was taken out in https://github.com/IBM/node-sdk-core/pull/6, and it currently is never read, so the change here should have no BC issues.

dpopp07:
@mkistler I agree

I'm waiting until the v5 release goes out and then will release some maintenance changes for v4. In the meantime, I'll bring it up with the other SDK owners to roll out this change after the major release

dpopp07:
Included in `v0.3.7`

PR #60, opened 2019-10-03
Release v1.0.0

dpopp07:
PR to merge all major release changes back into `master`. This will kick off the 1.0.0 release.

Note that all changes were previously reviewed when merged into the release branch.


auth/authenticators/basic-authenticator.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import extend = require('extend');
+import { computeBasicAuthHeader, validateInput } from '../utils';
+import { Authenticator } from './authenticator';
+import { AuthenticateOptions, AuthenticatorInterface } from './authenticator-interface';
+
+export type Options = {
+  username?: string;
+  password?: string;
```
mkistler: Why are username and password optional here?

dpopp07: Wow good catch - glad I asked for a review! That's a mistake and I will fix right now

dpopp07: fixed

lib/base-service.ts
```diff
@@ -0,0 +1,185 @@
...
+   * Wrapper around `sendRequest` that enforces the request will be authenticated.
+   *
+   * @param {Object} parameters - service request options passed in by user.
+   * @param {string} parameters.options.method - the http method.
+   * @param {string} parameters.options.url - the path portion of the URL to be appended to the serviceUrl
+   * @param {string} parameters.options.path - the path to be appended to the service URL.
+   * @param {string} parameters.options.qs - the querystring to be included in the URL.
+   * @param {string} parameters.options.body - the data to be sent as the request body.
+   * @param {Object} parameters.options.form - an object containing the key/value pairs for a www-form-urlencoded request.
+   * @param {Object} parameters.options.formData - an object containing the contents for a multipart/form-data request.
```
mkistler: The convention is that we designate optional parameters by putting them in brackets (`[]`), right?  If so, should we indicate this for `parameters.options.qs`, `parameters.options.body`, `parameters.options.form`, and `parameters.options.formData`?

dpopp07: Yes, that's a good call. I think `parameters.options.path` also. Those are path parameters - I will change the description as well.

dpopp07: fixed

PR #62, opened 2019-10-04
feat: adding winston logger

jorge-ibm:
Added a basic implementation of a logger through ``winston``. Log messages are suppressed by default unless the specified environment variables in the README are set. I've gone ahead and updated some unit tests that broke after this change, added unit tests for the logger itself, and replaced all ``console.log`` references throughout the code.

##### Checklist

- [x] tests are included
- [x] documentation is changed or added (please see README)


test/unit/logger.test.js
```diff
@@ -0,0 +1,251 @@
...
+    expect(removedTransport).toBeUndefined();
+  });
+
+  xit('should clear all transports', () => {
+    const newLogger = logger;
+    newLogger.clear();
+    const transports = newLogger.transports;
+
+    expect(transports.length).toEqual(0);
+  });
```
dpopp07: Why are these skipped?

README.md
```diff
@@ -146,3 +146,18 @@ import { getAuthenticatorFromEnvironment } from 'ibm-cloud-sdk-core';
...
+
+When disabled, the logger has the following configuration option:
+- `LOG_TO_CONSOLE` - set to `true` if you wish to log messages to the console, but not to a log file. This is disabled by default.
+
+When enabled, the logger has a set of configuration options:
+
+- `LOG_LEVEL` - severity level of logging as outlined in the [winston docs](https://github.com/winstonjs/winston#logging). This is set to `info` by default.
+- `LOG_FILE_NAME ` - path of where to write the logs to. By default this is set to the current working directory `./logs/node-sdk-core.log`
+- `ENABLE_ROTATING_FILE` - set to `true` if you wish to [enable rotating file support](https://github.com/winstonjs/winston-daily-rotate-file). If enabled, this will write to the file specified by `LOG_FILE_NAME`. This is disabled by default.
+- `LOG_TO_CONSOLE` - set to `true` if you wish to also log messages to the console. Enabling this while the logger is enabled will log both to a file and the console. This is disabled by default.
```
dpopp07: Reading through this again, there may be a simpler way to configure this. Can we do something like:
- Logger is off by default
- if `LOG_TO_CONSOLE` is defined, add a console transport
- if `LOG_FILE_NAME` is defined, add a file transport
- Rotating daily file and log level work the same way

That way we don't have to rely on the `ENABLE_LOGGER` flag as an extra variable the user needs to set. What do you think?

test/unit/logger.test.js
jorge-ibm: @dpopp07 - I had originally planned on giving the user full control of the logger configuration, so I wrote tests for it. I left them in there in case we wanted to go that route in the future, but I don't have any concerns if we just get rid of them.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -156,6 +158,7 @@ export class JwtTokenManager {
     const accessToken = tokenResponse[this.tokenName];
 
     if (!accessToken) {
+      logger.error('Access token not present in response');
       throw new Error('Access token not present in response');
```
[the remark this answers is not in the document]
jorge-ibm: I agree. That actually crossed my mind when I did this but didn't go that route. I'll go ahead and make a change for it.

test/unit/logger.test.js
jorge-ibm: Sure

README.md
jorge-ibm: I've gone ahead and updated the readme to follow something similar to what you suggested.

lib/logger.ts
```diff
@@ -0,0 +1,43 @@
...
+// If LOG_FILE_NAME is set, then log to specified file
+if (process.env.LOG_FILE_NAME && !process.env.ENABLE_ROTATING_FILE) {
+  fileTransport = new winston.transports.File({ filename: fileName });
+  logger.add(fileTransport);
+}
+// If flag is enabled then remove existing file transport & add rotating log file transport
+else if (process.env.ENABLE_ROTATING_FILE === 'true') {
+  if (fileTransport) {
+    logger.remove(fileTransport);
+  }
```
dpopp07: Correct me if I'm wrong, but this block seems like it will never execute thanks to the if/else that this is wrapped in so it probably isn't necessary

jorge-ibm: Nope, you are right. Good catch, I originally had it set to just two if statements

lib/request-wrapper.ts
```diff
@@ -69,40 +70,40 @@ export class RequestWrapper {
     // set debug interceptors
     if(process.env.NODE_DEBUG === 'axios') {
```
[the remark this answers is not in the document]
dpopp07: Ah yes, I like that

germanattanasio:
We should make sure the libraries we use are using Winston or we configure them with the logging environment variables.

Why [`winston`](https://www.npmjs.com/package/winston) rather than [`debug`](https://www.npmjs.com/package/debug)

mkistler:
Regarding `winston` vs `debug`, it's worth noting that Microsoft has chosen `debug` as their standard for Node/JS/TS logging:

https://azure.github.io/azure-sdk/typescript_implementation.html#general-logging

jorge-ibm:
Thanks @mkistler, @germanattanasio. This is the first time I hear about `debug`. Do we feel like that should be the chosen logging method? I went with winston since in my experience it's a pretty flexible logger and I have not had any issues with it. Though I do see `debug` is more lightweight, but similar in concept to what I have done with winston and the use of environment variables.

jorge-ibm:
As far as dependency logging, the core ones that @dpopp07 mentioned to me seem to not have logging enabled by us.

germanattanasio:
@jorge-ibm will the SDK work in the browser or react-native sort of environment? I think that's why Microsoft went with `debug`.

jorge-ibm:
@germanattanasio I want to say browser support was recently added, but correct me if I'm wrong @dpopp07 . If that's something we should consider, then yes I think winston would be an issue since it doesn't look like browser is supported _yet_ . (it's in their [roadmap](https://github.com/winstonjs/winston/blob/master/CONTRIBUTING.md#version-320).)

dpopp07:
We've always had browser support, what we did recently was add a way to prevent an issue with a Node-only module. I think we would have to add another patch like that for the logger if we go with `winston`. Maybe that is reason enough to consider going with `debug`, although `winston` looks more flexible overall.

jorge-ibm:
Sounds like debug is probably more safe to use then. I will look into implementing that instead of winston. Thanks everyone!

PR #64, opened 2019-10-08
feat: adding debug logger

jorge-ibm:
Based on feedback I have added logging through the _debug_ module. Log messages are suppressed by default unless the specified environment variables in the README are set. I've gone ahead and updated some unit tests that broke after this change, added unit tests for the logger itself, and replaced all ``console.log`` references throughout the code.

##### Checklist

- [x] tests are included

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,26 @@ import { getAuthenticatorFromEnvironment } from 'ibm-cloud-sdk-core';
 const iamAuthenticator = getAuthenticatorFromEnvironment('my-service');
 ```
 
+## Logging
+This package uses [debug](https://www.npmjs.com/package/debug) for logging.
+
+- Logging is disabled by default.
+- Logging has been configured to use log levels which are assumed to be numerically ascending from most important to least important.
+- In order to see the log output, set the environment variable ``DEBUG`` including the desired log level.
+  - ```DEBUG=ibm-cloud-sdk-core:error``` enables error logs
+  - ```DEBUG=ibm-cloud-sdk-core:warning``` enables warning logs and below
+  - ```DEBUG=ibm-cloud-sdk-core:info``` enables info logs and below
+  - ```DEBUG=ibm-cloud-sdk-core:verbose``` enables verbose logs and below
+  - ```DEBUG=ibm-cloud-sdk-core:debug``` enables debug logs and below
+
+To see the output from all of the debugging levels you can use:
+
+``DEBUG=ibm-cloud-sdk-core*``
+
+The debug logger can be configured to be used for more than one library. In example, you can set a comma-separated string:
+
+``DEBUG=ibm-cloud-sdk-core:debug,other-lib:debug``
+
 ## Issues
 If you encounter an issue with this project, you are welcome to submit a [bug report](https://github.com/IBM/node-sdk-core/issues).
 Before opening a new issue, please search for similar issues. It's possible that someone has already reported it.
--- a/auth/utils/read-credentials-file.ts
+++ b/auth/utils/read-credentials-file.ts
@@ -2,6 +2,7 @@ import dotenv = require('dotenv');
 import fs = require('fs');
 import os = require('os');
 import path = require('path');
+import logger from '../../lib/logger';
 
 const filename: string = 'ibm-credentials.env';
 
@@ -36,6 +37,7 @@ export function readCredentialsFile() {
     filepathToUse = homeDir;
   } else {
     // file does not exist anywhere, will not be used
+    logger.info('Credential file does not exist. Will not be used');
     return {};
   }
 
--- a/lib/base-service.ts
+++ b/lib/base-service.ts
@@ -21,6 +21,7 @@ import semver = require('semver');
 import vcapServices = require('vcap_services');
 import { AuthenticatorInterface, checkCredentials, readExternalSources } from '../auth';
 import { stripTrailingSlash } from './helper';
+import logger from './logger';
 import { RequestWrapper } from './request-wrapper';
 
 export interface UserOptions {
@@ -65,9 +66,9 @@ export class BaseService {
    */
   constructor(userOptions: UserOptions) {
     if (!(this instanceof BaseService)) {
-      throw new Error(
-        'the "new" keyword is required to create service instances'
-      );
+      const err = 'the "new" keyword is required to create service instances';
+      logger.error(`Error creating an instance of BaseService: ${err}`);
+      throw new Error(err);
     }
 
     const _options = {} as BaseServiceOptions;
@@ -85,6 +86,7 @@ export class BaseService {
     // check serviceUrl for common user errors
     const credentialProblems = checkCredentials(options, ['serviceUrl']);
     if (credentialProblems) {
+      logger.error(credentialProblems.message);
       throw credentialProblems;
     }
 
--- a/lib/logger.ts
+++ b/lib/logger.ts
@@ -1,0 +1,31 @@
+import logger = require('debug');
+const debug = logger('ibm-cloud-sdk-core:debug');
+const error = logger('ibm-cloud-sdk-core:error');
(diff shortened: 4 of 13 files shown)


lib/request-wrapper.ts
```diff
@@ -67,42 +68,42 @@ export class RequestWrapper {
...
       this.axiosInstance.interceptors.request.use(config => {
-        console.debug('Request:');
+        logger.debug('Request:');
         try {
-          console.debug(JSON.stringify(config, null, 2));
+          logger.debug(JSON.stringify(config, null, 2));
         } catch {
-          console.debug(config)
+          logger.error(config)
         }
```
mkistler: There's something still not quite right here.  If someone sets `NODE_DEBUG=axios` but _does not_ set `DEBUG=node-sdk-core:debug`, then they will get _no_ debug output (if I'm reading this correctly).

It appears that we don't actually document the behavior of `NODE_DEBUG=axios`, so I guess we need to decide how we want this to work.  We could just drop support for `NODE_DEBUG=axios` now that we have this new logging support.

jorge-ibm: Ah yes you're right. Was that flag initially added just to log the request and response being intercepted?

README.md
```diff
@@ -59,6 +59,21 @@ import { getAuthenticatorFromEnvironment } from 'ibm-cloud-sdk-core';
...
+  - ```DEBUG=node-sdk-core:error``` enables error logs
+  - ```DEBUG=node-sdk-core:info``` enables info logs
+  - ```DEBUG=node-sdk-core:debug``` enables debug logs
+  - ```DEBUG=node-sdk-core:verbose``` enables verbose logs
+  - ```DEBUG=node-sdk-core:warning``` enables warning logs
+
+To see the output from all of the debugging scopes you can use:
+
+```DEBUG=node-sdk-core*```
+
```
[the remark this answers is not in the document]
jorge-ibm: So the way I set the logger up is for all the debug instances to be under ``node-sdk-core``. So the user can only enable/disable loggers under one library ``DEBUG=node-sdk-core:debug,node-sdk-core:error``. I originally did have a logger for each library but moved towards the simpler approach of just having one for each log scope. Do we want to have my original approach instead? One for baseservice, helper, etc.. ?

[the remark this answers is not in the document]
jorge-ibm: @mkistler - please see the changes I made to the README. Let me know if this doesn't work.

lib/request-wrapper.ts
jorge-ibm: @mkistler - I removed the log flag from here and I just enable all the loggers if the axios flag is set. Please take a look at ``logger.ts``

jorge-ibm: ~@mkistler - I removed the DEBUG flag from here and I just enabled all the loggers if the axios flag is set. Please take a look at `logger.ts`~ I just realized removing ``DEBUG`` flag was a mistake. Since removing made that block dependent on the axios flag being set now. Re-adding it back in.

germanattanasio:
This is looking good @jorge-ibm Buen trabajo

README.md
```diff
@@ -59,6 +59,25 @@ import { getAuthenticatorFromEnvironment } from 'ibm-cloud-sdk-core';
...
+## Logging
+This package uses [debug](https://www.npmjs.com/package/debug) for logging.
+
+- Logging is disabled by default.
+- In order to see the log output, set the environment variable ``DEBUG`` including the desired scope.
+  - ```DEBUG=ibm-cloud-sdk-core:error``` enables error logs
+  - ```DEBUG=ibm-cloud-sdk-core:info``` enables info logs
+  - ```DEBUG=ibm-cloud-sdk-core:debug``` enables debug logs
+  - ```DEBUG=ibm-cloud-sdk-core:verbose``` enables verbose logs
+  - ```DEBUG=ibm-cloud-sdk-core:warning``` enables warning logs
```
dpopp07: I suppose this is fine but I feel like the typical idea in loggers is that you set _one_ level and then you get the logs from that level and also everything "under" that level in terms of decreasing verbosity. For example, if I set the level "verbose", I would also expect to get error and warnings and info, etc.

This seems like a limitation of `debug` and not your implementation so I'm not sure there's anything we can do here but I hope we aren't creating an experience that's less than idea

jorge-ibm: This wasn't specified in the documentation, and yes it's a limitation of using debug. We can manually force this ourselves by enabling each log level through some conditions, but something to keep in mind is that I noticed debug was being used by other packages in our code. So that might make our version of log levels different than what the other packages do if they did not manually enforce that.

mkistler: I noticed the same thing and would like to put in my vote for making the levels "additive", with a clear indication of the order of inclusion.

lib/request-wrapper.ts
```diff
@@ -67,42 +68,42 @@ export class RequestWrapper {
...
+          logger.error(response);
         }
 
         return response;
       }, error => {
-        console.debug('Error:');
+        logger.error('Error returning response: ');
         try {
-          console.debug(JSON.stringify(error, null, 2));
+          logger.error(JSON.stringify(error, null, 2));
```
mkistler: If the error logged here does not include the original request, it would be really nice if we could find a way to log the request here as well.

jorge-ibm: So I ran a test and it looks like the error object returned does in fact contain the original request (under ``error.config``), so as long as the stringify has been doing its job then that should have been logged as well

README.md
jorge-ibm: Sounds good. I've gone ahead and implemented log levels

lib/logger.ts
```diff
@@ -0,0 +1,35 @@
...
+const verbose = logger('ibm-cloud-sdk-core:verbose');
+const warn = logger('ibm-cloud-sdk-core:warning');
+
+// enable loggers if axios flag is set & mimic log levels severity
+if (process.env.NODE_DEBUG === 'axios') {
+  debug.enabled = true;
+  verbose.enabled = true;
+  info.enabled = true;
+  warn.enabled = true;
+  error.enabled = true;
```
mkistler: You don't need these four lines.  The logic below handles this.

README.md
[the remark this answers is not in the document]
germanattanasio: What's the difference between `verbose` and `debug`?

auth/utils/read-credentials-file.ts
```diff
@@ -2,11 +2,13 @@ import dotenv = require('dotenv');
 import fs = require('fs');
 import os = require('os');
 import path = require('path');
+import logger from '../../lib/logger';
 
 const filename: string = 'ibm-credentials.env';
 
 export function readCredentialsFile() {
   if (!fs.existsSync) {
+    logger.warn('Unable to find the library fs.existsSync');
```
germanattanasio: This usually happen when running in a browser and it's not a warning. I can't come up with a better message but I see this from a browser developer perspective and that person will be getting a warning that is not a problem

dpopp07: I'll add a +1 here, we can probably just remove that log

lib/base-service.ts
```diff
@@ -65,6 +66,7 @@ export class BaseService {
    */
   constructor(userOptions: UserOptions) {
     if (!(this instanceof BaseService)) {
+      logger.error('Error creating an instance of BaseService: the "new" keyword is required to create service instances');
       throw new Error(
         'the "new" keyword is required to create service instances'
```
dpopp07: This is picky but the string `the "new" keyword is required to create service instances` _could_ be assigned to a variable since it is used twice here. Not a required change

README.md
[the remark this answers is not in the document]
jorge-ibm: Debug level includes all the logs, and verbose does as well excluding debug. This mimics winston's log levels based on the standard https://tools.ietf.org/html/rfc5424

PR #65, opened 2019-10-10
feat: export unit test utility methods to be used in SDKs

dpopp07:
* these are not for the core tests themselves - the generated unit tests for service client sdks rely on these methods

Right now, we just define these in the SDK itself. They are also included in the template repo. It doesn't make sense to keep all of these different files in sync across the SDKs once new services start using them. If the generated code/tests change, this file may need to change. We need to control this code in a central location.

Usage would look like this:

```js
const { unitTestUtils } = require('ibm-cloud-sdk-core');

const {
  getOptions,
  checkUrlAndMethod,
  checkMediaHeaders,
  expectToBePromise,
  checkDefaultSuccessArgs,
} = unitTestUtils;
```

I also wanted to note that I structured the code the way I did to easily allow more test utility methods in future (like helper methods for generated integration tests).


test/utils/unit-test-helpers.js
```diff
@@ -0,0 +1,83 @@
...
+
+/* istanbul ignore file */
+
+const missingParamsError = 1;
+const missingParamsSuccess = 0;
+
+module.exports.missingParamsError = missingParamsError;
+module.exports.missingParamsSuccess = missingParamsSuccess;
+
+module.exports.checkUrlAndMethod = function(options, url, method) {
```
mkistler: I think we should have a description on any method that is exported from the package.  This doesn't need to be elaborate -- maybe just a one sentence statement of behavior.

test/utils/unit-test-helpers.js
```diff
@@ -0,0 +1,83 @@
...
+  expect(options.url).toEqual(url);
+  expect(options.method).toEqual(method);
+};
+
+module.exports.checkMediaHeaders = function(createRequestMock, accept, contentType) {
```
mkistler: Why is the first parameter called "createRequestMock"?  Is this method only intended to be used for create requests?

test/utils/unit-test-helpers.js
```diff
@@ -0,0 +1,83 @@
...
+  expect(headers[baseName]).toEqual(param);
+};
+
+module.exports.checkDefaultSuccessArgs = function(createRequestMock) {
+  // get arg to getMissingParams
+  const userParams = createRequestMock.mock.calls[0];
+  expect(typeof userParams[0]).toEqual('object');
+};
+
+module.exports.checkForEmptyObject = function(missingParamsMock) {
```
mkistler: This method needs a more descriptive name.  Check what for empty object?  Maybe "checkUserParamsIsEmptyObject".

test/utils/unit-test-helpers.js
```diff
@@ -0,0 +1,83 @@
...
+  // empty object should always be used as params
+  const params = {};
+
+  // assert getMissingParams was called and extract called arguments
+  expect(mpMock).toHaveBeenCalledTimes(1);
+  const userParams = mpMock.mock.calls[0][0];
+  const validatorParams = mpMock.mock.calls[0][1];
+
+  // assert getMissingParams is called with correct args
+  expect(userParams).toEqual(params);
```
mkistler: Seems like we could just use our method above to perform this check.

test/utils/unit-test-helpers.js
```diff
@@ -0,0 +1,83 @@
...
+  expect(validatorParams).toEqual(required);
+
+  // assert callback is called with missingParamsError
+  expect(err).toEqual(missingParamsError);
+
+  // assert createRequest is never called
+  expect(crMock).toHaveBeenCalledTimes(0);
```
mkistler: Oh ... I guess `crMock` is "createRequest Mock".  That was not obvious to me until just now.  I think we should use more descriptive parameter names.

test/utils/unit-test-helpers.js
dpopp07: Agreed, I will add that

test/utils/unit-test-helpers.js
dpopp07: Ah, you are correct. Good catch

dpopp07:
@mkistler All feedback addressed. In-code documentation added for each test helper. I will be making corresponding changes to my generator PR as well

test/utils/unit-test-helpers.js
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Verifies that when a user does not provide the required parameters for a method, the validation step casues the method to fail.
+ * The `getMissingParams` mock function is expected to return an error and the code is not expected to ever call the `createRequest`
+ * method.
+ *
+ * @param {Object} missingParamsMock - the jest mock object for the `getMissingParams` function in  the `helpers` module
+ * @returns {void}
+ */
+module.exports.checkRequiredParamsHandling = function(
```
mkistler: This method is still confusing.  I was really only able to understand what it was doing by looking at the unit tests to see how they were using it.

Now that I do understand it, I think we probably want to change our unit tests so that we don't need this method.  I don't think we should mock `getMissingParameters` at all -- we can completely observe its behavior by inspecting the exception that is returned.  I'll elaborate this in the generator PR, but if we agree on that new approach, then we can simply remove this method from the core.

dpopp07:
Feedback addressed! The three methods that were rendered unnecessary by your generator PR have been removed.

PR #66, opened 2019-10-22
feat: adding configureService method for external config options

jorge-ibm:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/1118

Changes:
- moved the external configuration code outside of the constructor into it's own method.
- for compatibility, the constructor calls the new method until we move to the service factory's complete implementation.

The new method will now be invoked from the generator after the BaseService's constructor is called.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/base-service.ts
+++ b/lib/base-service.ts
@@ -100,7 +100,6 @@ export class BaseService {
     this.baseOptions = extend(
       { qs: {}, serviceUrl: serviceClass.URL },
       options,
-      this.readOptionsFromExternalConfig(),
       _options
     );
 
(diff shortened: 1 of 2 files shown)


lib/base-service.ts
```diff
@@ -132,6 +134,27 @@ export class BaseService {
     this.baseOptions.serviceUrl = url;
   }
 
+  /**
+   * Configure the service using external configuration
+   *
+   * @param {string} the name of the service. Will be used to read from external
+   * configuration
+   */
+  protected configureService(serviceName: string): void {
```
dpopp07: I can't remember - is the idea to have users call this in the future? If so, this should be public. If not, this is good

dpopp07: After discussing with @jorge-ibm and @padamstx, the intention is for this method to only be called from generated code, so `protected` is fine 👍

PR #72, opened 2019-11-14
feat: changes to node-sdk-core to work with service factory feature

SaltedCaramelCoffee:
##### Checklist
- [x] tests are included


auth/utils/read-external-sources.ts
```diff
@@ -97,8 +96,31 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
  * Pulls credentials from VCAP_SERVICES env property that IBM Cloud sets
  *
  */
+function getCredentials(name) {
```
padamstx: After copying this code from its original package into this file, I think it would be good to give this a more descriptive name.  
In fact, I'll suggest that you rename `getCredentialsFromCloud` to be `getPropertiesFromVCAP` and then rename `getCredentials` to be something like `getVCAPCredentialsForService`, or just move the `getCredentials` logic into the newly-renamed `getPropertiesFromVCAP`.

test/unit/getCredentials.test.js
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it('should return empty credential object when looking for credentials by service name', () => {
+    const credentials = readExternalSources('discovery1')
+    expect(credentials.url).toEqual("https://gateway.watsonplatform1.net/discovery-experimental/api");
+    expect(credentials.username).toEqual("not-a-username")
+    expect(credentials.password).toEqual("not-a-password")
+  });
+
+  it('should return first service in the list when multiple services found', () => {
```
padamstx:
```suggestion
  it('should return first service in the list when matching on primary service key', () => {
```

test/unit/getCredentials.test.js
```diff
@@ -0,0 +1,45 @@
...
+    process.env['VCAP_SERVICES'] = fs.readFileSync(__dirname + '/../resources/vcap.json')
+  })
+
+  it('should return empty credential object when service key is empty_service', () => {
+    const credentials = readExternalSources('empty_service')
+    expect(credentials).toBeDefined()
+    expect(Object.keys(credentials).length).toEqual(0);
+  });
+
+  it('should return empty credential object when looking for credentials by service name', () => {
```
padamstx:
```suggestion
  it('should return credential object matching service name field not first list element', () => {
```

test/unit/getCredentials.test.js
```diff
@@ -0,0 +1,45 @@
...
+    const credentials = readExternalSources('discovery2')
+    expect(credentials.url).toEqual("https://gateway.watsonplatform2.net/discovery-experimental/api");
+    expect(credentials.username).toEqual("not-a-username")
+    expect(credentials.password).toEqual("not-a-password")
+  });
+
+  it('should return empty credential object when looking for credentials by service name', () => {
```
padamstx:
```suggestion
  it('should return credential object when matching service name field on first list element', () => {
```

test/unit/getCredentials.test.js
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it('should return first service in the list when multiple services found', () => {
+    const credentials = readExternalSources('discovery')
+    expect(credentials.url).toEqual("https://gateway.watsonplatform1.net/discovery-experimental/api");
+    expect(credentials.username).toEqual("not-a-username")
+    expect(credentials.password).toEqual("not-a-password")
+  });
+
+  it('should return empty credential object when service key is no-creds-service-two', () => {
```
padamstx:
```suggestion
  it('should return empty credential object when matching on service name with no credentials field', () => {
```

auth/utils/read-external-sources.ts
```diff
@@ -97,8 +96,31 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
  * Pulls credentials from VCAP_SERVICES env property that IBM Cloud sets
  *
  */
-function getCredentialsFromCloud(serviceName: string): any {
-  const credentials = vcapServices.getCredentials(serviceName);
+function getVCAPCredentialsForService(name) {
+  if (process.env.VCAP_SERVICES) {
+    const services = JSON.parse(process.env.VCAP_SERVICES);
+    for (const serviceName of Object.keys(services)) {
```
dpopp07: > for (const serviceName of Object.keys(services)) {

Isn't this the same thing as `for (const serviceName in services) {` below?

It might be more concise to combine the loops into something like:

```js
    for (const serviceName in services) {
      const service = services[serviceName];
      for (const instance of service) {
        if (instance.name === name) {
          return instance.credentials || {};
        }
      }
      if (serviceName === name) {
        if (service.length > 0) {
          return service[0].credentials || {};
        } else {
          return {}
        }
      }
    }
```

I just found myself as the reader a little confused why there were two different loops for the same thing. This snippet includes some other things to make the code a little more concise as well.

auth/utils/read-external-sources.ts
```diff
@@ -97,8 +96,31 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
  * Pulls credentials from VCAP_SERVICES env property that IBM Cloud sets
  *
  */
-function getCredentialsFromCloud(serviceName: string): any {
-  const credentials = vcapServices.getCredentials(serviceName);
+function getVCAPCredentialsForService(name) {
+  if (process.env.VCAP_SERVICES) {
+    const services = JSON.parse(process.env.VCAP_SERVICES);
```
dpopp07: I think it would be safer to wrap the JSON parse in a `try/catch` since it throws an error if anything about the JSON is malformatted. Not that it should be, since that env variable is generated by the cloud. So this is an optional change, up to you.

```suggestion
    let services;
    try {
      services = JSON.parse(process.env.VCAP_SERVICES);
    } catch () {
      return {};
    }
```

auth/utils/read-external-sources.ts
```diff
@@ -97,8 +96,31 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
  * Pulls credentials from VCAP_SERVICES env property that IBM Cloud sets
  *
  */
-function getCredentialsFromCloud(serviceName: string): any {
-  const credentials = vcapServices.getCredentials(serviceName);
+function getVCAPCredentialsForService(name) {
```
dpopp07: My only other suggestion is that it would be useful to have a comment describing what this algorithm is looking for. Like "looks for an exact match of the service name, otherwise defaults to the first whatever, blah, blah"

Describing the intent will make the code more readable I think.

auth/utils/read-external-sources.ts
```diff
@@ -16,7 +16,6 @@
 
 import camelcase = require('camelcase');
 import isEmpty = require('lodash.isempty');
-import vcapServices = require('vcap_services');
```
dpopp07: If we're removing all use of this package, it should be removed from the `package.json` as well.

lib/base-service.ts
```diff
@@ -123,9 +121,6 @@ export class BaseService {
     }
 
     this.authenticator = options.authenticator;
-
-    // temp: call the configureService method to ensure compatibility
-    this.configureService(this.name);
```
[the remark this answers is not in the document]
padamstx: Yes, which I think means that the commit message will need to be updated to make that happen, right?

auth/utils/read-external-sources.ts
dpopp07: Okay actually, there is an extremely unlikely edge case where my code would not work (if one of the top level names matches a specific "name" field under one of the other top level services. So the two loops are fine, but I would still use `for (const serviceName in services) {` for both for clarity.

auth/utils/read-external-sources.ts
SaltedCaramelCoffee: Wrapping the JSON parse in a `try/catch` might eat up/silence the error, and the user wouldn't know their `VCAP_SERVICES` env is not parsed. So I think it's fine to let it fail if for some reason JSON is malformatted.

auth/utils/read-external-sources.ts
SaltedCaramelCoffee: `for (const serviceName in services)` failed tslint with the error: `read-external-sources.ts[102, 5]: for (... in ...) statements must be filtered with an if statement`. But you're right in that the 2 for-loops does the same thing, but written differently. I'll changed the 2nd for-loop to match the first one to make it consistent.

auth/utils/read-external-sources.ts
dpopp07: Okay, that's a fair point. I just remembered that the core has it's own logger now. So another thing to consider would be printing an error log for a malformatted VCAP variable so the user knows about it. My only concern here is the code crashing even though the user typically cannot control the contents of `VCAP_SERVICES`.

auth/utils/read-external-sources.ts
dpopp07: Huh, that's a weird linting rule, but fair enough. That sounds good

lib/base-service.ts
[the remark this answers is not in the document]
dpopp07: That's right. @SaltedCaramelCoffee we need to make sure the final commit message is formatted as such:
```
feat: whatever whatever

BREAKING CHANGE: The code will no longer read configuration data from external sources by default (or something like this)
```

auth/utils/read-external-sources.ts
```diff
@@ -97,8 +96,31 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
...
+          return instance.credentials || {};
+        }
+      }
+    }
+    for (const serviceName in services) {
+      if (serviceName === name) {
+        if (services[serviceName].length > 0) {
+          return services[serviceName][0].credentials || {};
+        } else {
+          return {}
```
dpopp07: Speaking of the logger, I think it's a good idea to add debug or verbose logging statements before an empty object is returned saying something like "no data read from vcap services"

auth/utils/read-external-sources.ts
```diff
@@ -96,9 +96,50 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
...
+          }
+          return services[serviceName][0].credentials || {};
+        } else {
+          logger.debug('no data read from VCAP_SERVICES')
+          return {}
+        }
+      }
+    }
+  }
+  return {};
```
dpopp07: One more place to add a debug

auth/utils/read-external-sources.ts
```diff
@@ -96,9 +96,50 @@ function filterPropertiesByServiceName(envObj: any, serviceName: string): any {
 /**
  * Pulls credentials from VCAP_SERVICES env property that IBM Cloud sets
  *
+ * The function will first look for a service entry whose "name" field matches
+ * the serviceKey value. If found, return its credentials.
+ *
+ * If no match against the service entry's "name" field is found, then find the
+ * service list with a key matching the serviceKey value. If found, return its
+ * credentials.
```
dpopp07: Should mention that you are returning the first credentials object in the service list.

PR #78, opened 2020-01-29
fix: use consistent fields between success and error response objects

dpopp07:
The fields `status` and `statusText` are now added to the response error, to be consistent with the fields on a successful response object. The other fields are left intact for compatibility but are marked as deprecated so we know to remove them in the next major release.


dpopp07:
That's a good idea. I'll open an issue for that on our board.

PR #79, opened 2020-02-01
Pace token refresh requests to avoid rate-limiting

mkistler:
This PR modifies the token refresh logic to pace refresh requests to avoid exceeding rate-limiting by the token server.  This design also improves performance by continuing to use an existing token while the refresh is performed concurrently.

I also cleaned up some misinformation in the `CONTRIBUTING.md` that I uncovered in the process of developing this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,30 +25,24 @@ If you want to contribute to the repository, follow these steps:
 1. Fork the repo.
 2. Install dependencies: `npm install`
 3. Build the code: `npm run build`
-4. Verify the build before beginning your changes: `npm run test-unit`
+4. Verify the build before beginning your changes: `npm test`
 2. Develop and test your code changes.
-3. Travis-CI will run the tests for all services once your changes are merged.
 4. Add a test for your changes. Only refactoring and documentation changes require no new tests.
-5. Make the test pass.
+5. Make sure all the tests pass.
+2. Check code style with `npm run lint` and fix any issues.
 6. Commit your changes. Remember the follow the correct commit message guidelines.
 7. Push to your fork and submit a pull request.
+3. Travis-CI will run the tests once your changes are merged.
 8. Be sure to sign the CLA.
 
 ## Tests
 
-Out of the box, `npm test` runs linting, unit tests, and integration tests (which require credentials).
-To run only the unit tests (sufficient for most cases), use `npm run test-unit`.
-
-To run the integration tests, you need to provide credentials to the integration test framework.
-The integration test framework will skip integration tests for any service that does not have credentials,
-
-To provide credentials for the integration tests, copy `test/resources/auth.example.js` to `test/resources/auth.js`
-and fill in credentials for the service(s) you wish to test.
+`npm test` runs the full suite of unit tests.
 
 To run only specific tests, invoke jest with the `--testNamePattern` or `-t` flag:
 
 ```
-npm run jest -- /test/integration -t "resources key"
+npm run jest -- /test/unit -t "resources key"
 ```
 
 This will only run tests that match the test name pattern you provide.
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -53,6 +53,7 @@ export class JwtTokenManager {
   protected requestWrapperInstance: RequestWrapper;
   private tokenInfo: any;
   private expireTime: number;
+  private refreshTime: number;
 
   /**
    * Create a new [[JwtTokenManager]] instance.
@@ -97,6 +98,12 @@ export class JwtTokenManager {
         return this.tokenInfo[this.tokenName];
       });
     } else {
+      // If refresh needed, kick one off
+      if (this.tokenNeedsRefresh()) {
+        this.requestToken().then(tokenResponse => {
+          this.saveTokenInfo(tokenResponse.result);
+        });
+      }
       // 2. use valid, managed token
       return Promise.resolve(this.tokenInfo[this.tokenName]);
     }
@@ -142,8 +149,7 @@ export class JwtTokenManager {
   }
 
   /**
-   * Check if currently stored token is "expired"
-   * i.e. past the window to request a new token
+   * Check if currently stored token is expired
    *
    * @private
    * @returns {boolean}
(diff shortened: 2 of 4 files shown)


auth/token-managers/jwt-token-manager.ts
```diff
@@ -90,13 +90,19 @@ export class JwtTokenManager {
    *   has expired.
    */
   public getToken(): Promise<any> {
-    if (!this.tokenInfo[this.tokenName] || this.isTokenExpired()) {
+    if (!this.tokenInfo[this.tokenName]) {
```
ricellis: I agree with @dpopp07 - I think the expiry check needs to remain here so that a synchronous request for a new token can be made in the case that a token's validity has lapsed after a period with no requests to force the preemptive refresh.
I guess you could add a test for this with a very short expiry time and a short block in a test case.

CONTRIBUTING.md
```diff
@@ -25,25 +25,19 @@ If you want to contribute to the repository, follow these steps:
...
 
-Out of the box, `npm test` runs linting, unit tests, and integration tests (which require credentials).
-To run only the unit tests (sufficient for most cases), use `npm run test-unit`.
-
-To run the integration tests, you need to provide credentials to the integration test framework.
-The integration test framework will skip integration tests for any service that does not have credentials,
-
-To provide credentials for the integration tests, copy `test/resources/auth.example.js` to `test/resources/auth.js`
-and fill in credentials for the service(s) you wish to test.
+`npm test` runs the full suite of unit tests.
```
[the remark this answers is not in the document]
mkistler: Fixed.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -175,36 +202,27 @@ export class JwtTokenManager {
...
-   *
-   * A time buffer prevents the edge case of the token expiring before the request could be made.
-   * The buffer will be a fraction of the total time to live - we are using 80%
-   *
-   * @param accessToken - JSON Web Token received from the service
-   * @private
-   * @returns {void}
-   */
-  private calculateTimeForNewToken(accessToken): number {
     // the time of expiration is found by decoding the JWT access token
```
mkistler: I had to integrate `calculateTimeForNewToken` into `saveTokenInfo` because we now need to set both the `expireTime` and `refreshTime` from the values in the decoded token.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -175,36 +202,27 @@ export class JwtTokenManager {
...
+    if (!decodedResponse) {
       const err = 'Access token recieved is not a valid JWT'
       logger.error(err);
       throw new Error(err);
     }
 
-    return timeForNewToken;
+    const { exp, iat } = decodedResponse;
+    // There are no required claims in JWT
+    if (!exp || !iat) {
```
mkistler: This is just extra bullet-proofing for the case where the JWT does not have an `exp` or `iat` claim.

test/unit/iam-token-manager.test.js
```diff
@@ -62,14 +63,11 @@ describe('iam_token_manager_v1', function() {
     const requestMock = jest.spyOn(instance, 'requestToken');
 
     const currentTokenInfo = {
-      access_token: '1234',
-      refresh_token: '5678',
-      token_type: 'Bearer',
-      expires_in: 3600,
-      expiration: Math.floor(Date.now() / 1000),
+      access_token: CURRENT_ACCESS_TOKEN,
```
mkistler: Since the `iam-token-manager` was changed to simply treat the access token as a JWT, none of the other fields in the IAM response are used, so we can eliminate them.

test/unit/iam-token-manager.test.js
```diff
@@ -79,64 +77,48 @@ describe('iam_token_manager_v1', function() {
...
-
-    const token = await instance.getToken();
-    expect(token).toBe(ACCESS_TOKEN);
-    expect(requestMock).not.toHaveBeenCalled();
-    done();
-  });
+    instance.expireTime = Math.floor(Date.now() / 1000) + 60;
+    instance.refreshTime = Math.floor(Date.now() / 1000) - 1;
 
-  it('should refresh an access token without expires_in field', async done => {
```
mkistler: This test was actually no longer valid, since the `expires_in` field is not used. It only happened to pass because the test did not set the `expireTime` in the instance.

test/unit/jwt-token-manager.test.js
```diff
@@ -33,7 +34,7 @@ describe('JWT Token Manager', () => {
 
       const decodeSpy = jest
         .spyOn(jwt, 'decode')
-        .mockImplementation(token => ({ iat: 0, exp: 100 }));
+        .mockImplementation(token => ({ iat: 10, exp: 100 }));
```
mkistler: Can't use `iat = 0` because that triggers the new check in the `jwt-token-manager`.

mkistler:
I've updated the code to handle an expired token the same a missing token -- with a synchronous request for a new token.  This turned out to be more invasive that I expected, but I think the new code now works as intended.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -159,6 +165,27 @@ export class JwtTokenManager {
     return expireTime < currentTime;
```
ricellis: Maybe this should really be `<=`. I know the chances of landing on it are remote, but if it does happen even though the token isn't quite expired yet I think there is basically zero chance the request is going to successfully reach the server & PDP within the same millisecond, so we probably may as well do the renewal and avoid the failure.

test/unit/jwt-token-manager.test.js
```diff
@@ -175,19 +177,40 @@ describe('JWT Token Manager', () => {
     });
   });
 
+  describe('tokenNeedsRefresh', () => {
+    it('should return true if current time is past expire time', () => {
```
ricellis: I think the test description is a bit misleading as it talks about "expire time", but in practice this is checking the `refreshTime`.

test/unit/jwt-token-manager.test.js
```diff
@@ -175,19 +177,40 @@ describe('JWT Token Manager', () => {
...
+      const instance = new JwtTokenManager();
+      instance.refreshTime = getCurrentTime() - 1000;
+
+      expect(instance.tokenNeedsRefresh()).toBe(true);
+    });
+
+    it('should return false if current time has not reached expire time', () => {
```
ricellis: expire -> refresh

test/unit/jwt-token-manager.test.js
```diff
@@ -175,19 +177,40 @@ describe('JWT Token Manager', () => {
...
+      const instance = new JwtTokenManager();
+      instance.refreshTime = getCurrentTime() + 1000;
+
+      expect(instance.tokenNeedsRefresh()).toBe(false);
+    });
+
+    it('should return true if expire time has not been set', () => {
```
ricellis: I think this one is ok, because if the expire time hasn't been set the refresh time doesn't get set.

PR #84, opened 2020-03-04
feat: optional cookie jar support

christiancompton:
This adds cookie jar support to Axios using [axios-cookiejar-support](https://github.com/3846masa/axios-cookiejar-support), so that consumers of the core can provide their own cookie jar management capabilities depending on the needs of their users.


test/unit/cookiejar.test.js
```diff
@@ -0,0 +1,40 @@
...
+  response: {
+    use: jest.fn(),
+  },
+};
+axios.default.create.mockReturnValue(mockAxiosInstance);
+
+const { RequestWrapper } = require('../../dist/lib/request-wrapper');
+const requestWrapperInstance = new RequestWrapper({ jar: new tough.CookieJar() });
+
+describe('axios', () => {
```
padamstx: Should we have a test or two that exercises cookie support?  i.e. we invoke a request, the "endpoint" returns a cookie of some sort in the response, and we then invoke another request and then verify that the cookie is in fact received by the "endpoint" when it receives the second request?

test/unit/cookiejar.test.js
```diff
@@ -0,0 +1,40 @@
...
+    env = process.env;
+    process.env = {};
+  });
+  afterEach(function() {
+    process.env = env;
+  });
+  it('should enable debug', () => {
+    // these should have been called when requestWrapperInstance and when the cookiejar wrapper was instantiated
+    expect(mockAxiosInstance.interceptors.request.use).toHaveBeenCalledTimes(2);
+    expect(mockAxiosInstance.interceptors.response.use).toHaveBeenCalledTimes(2);
```
dpopp07: What is the idea behind these tests? It seems like the interceptors and the cookie support should be unrelated

README.md
```diff
@@ -79,6 +79,22 @@ The debug logger can be configured to be used for more than one library. In exam
...
+
+class MyClass extends BaseService {
+  constructor(options: MyOptions) {
+    options.jar = new tough.CookieJar();
+    super(options)
+  }
+}
+```
+
+The example above uses [Tough Cookie](https://www.npmjs.com/package/tough-cookie) to provide these capabilities, but other cookie jar libraries can be used. Alternatively providing `jar: true`, will cause Axios to [instantiate a cookie jar](https://github.com/axios/axios#request-config).
```
dpopp07: I don't really think we should link the `axios` config after this statement - `jar` is not a documented option. I think it's a little confusing.

README.md
```diff
@@ -79,6 +79,22 @@ The debug logger can be configured to be used for more than one library. In exam
...
+```typescript
+import tough = require('tough-cookie');
+
+class MyClass extends BaseService {
+  constructor(options: MyOptions) {
+    options.jar = new tough.CookieJar();
+    super(options)
+  }
+}
+```
```
dpopp07: This documentation feels a little weird to me. SDK developers shouldn't be putting this in their SDK - it should be set by the end user. This documentation is for the SDK developer, so maybe it is helpful to show them how it would work. Perhaps, in this example, creating a new instance of `MyClass` and passing in the cookie as an option.

We eventually want the actual documentation to go into to SDK projects themselves, aka the template.

README.md
christiancompton: I am also confused by this...  The cookie wrapper library references that axios does this...  but I do not see anything in the Axios side that can do this. Per  https://github.com/3846masa/axios-cookiejar-support#extended-request-config

test/unit/cookiejar.test.js
christiancompton: One thing that happens when we wrap the instance the second time is that the interceptors are called twice - one on the initial wrapping and further on the  cookie wrapping.  This is not essential, but it is a good test that the instance is wrapped a second time.

README.md
christiancompton: @mkistler and I were talking about this - ultimately cookie jar support is a service-level decision.  Cloudant knows that its users who are on legacy services need this to support their basic authentication, so it makes sense to provide this functionality by default (if they choose) in their SDK, since all users will need to do it.

christiancompton: Putting it in the template is not a bad idea, but in the case where the SDK authors automatically configure the cookie jar, it would not be useful.

dpopp07: I see. I suppose that makes sense but it that's the case, does there need to be an optional cookie flag in the generator?

Also, there's nothing stopping any user creating a cookie jar at any time with this change. Do we want it to be a "secret" feature (undocumented) unless the service team enables it on a global level automatically?

I just want to make sure I fully understand our approach here

test/unit/cookiejar.test.js
dpopp07: Okay, I think I see where you're coming from. Will you update the comment you left to explain that a little more? I was confused on first read

README.md
dpopp07: I interpret that section of their README to mean that wrapping the `axios` with their tool provides an extended request configuration with those new options. I think it's clear enough to say what you're saying without attaching any links to it.

PR #88, opened 2020-04-08
expose the `saveTokenInfo` to subclasses

dpopp07:
This changes the exposure level of the `saveTokenInfo` method to be "protected" instead of "private" so that developers implementing a subclass can override that method as needed.

Additionally, this changes `saveTokenInfo` to accept the entire response and extract the body within the method. That way, there can be subclass implementations for authentication mechanisms that include the token somewhere other than the body, like the headers.

Reviewers: let me know if you think this is a breaking change. I decided that it was not because the only "contract" change is happening for `saveTokenInfo` (the argument changed) but that method was previously marked as "private". The `getToken` method was changed but the public contract is not affected by the change. I may be missing something though.

### DIFF
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -192,7 +192,7 @@ export class JwtTokenManager {
    * Save the JWT service response and the calculated expiration time to the object's state.
    *
    * @param tokenResponse - Response object from JWT service request
-   * @private
+   * @protected
    * @returns {void}
    */
   protected saveTokenInfo(tokenResponse): void {
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -105,7 +105,7 @@ export class JwtTokenManager {
       // If refresh needed, kick one off
       if (this.tokenNeedsRefresh()) {
         this.requestToken().then(tokenResponse => {
-          this.saveTokenInfo(tokenResponse.result);
+          this.saveTokenInfo(tokenResponse);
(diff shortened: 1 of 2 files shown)


auth/token-managers/jwt-token-manager.ts
```diff
@@ -188,6 +188,47 @@ export class JwtTokenManager {
...
+      this.refreshTime = 0;
+    } else {
+      const fractionOfTtl = 0.8;
+      const timeToLive = exp - iat;
+      this.expireTime = exp;
+      this.refreshTime = exp - (timeToLive * (1.0 - fractionOfTtl));
+    }
+
+    this.tokenInfo = extend({}, responseBody);
+  }
```
dpopp07, line 230: The methods are exactly the same, with the exception of this line: `const responseBody = tokenResponse.result || {};`, which was added. 

TS style requires the methods be listed in order of public -> protected -> private so I had to move the method, resulting in a larger-looking diff.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -188,6 +188,47 @@ export class JwtTokenManager {
     return Promise.reject(err);
   }
 
+  /**
+   * Save the JWT service response and the calculated expiration time to the object's state.
+   *
+   * @param tokenResponse - Response object from JWT service request
+   * @private
```
vmatyusGitHub, line 195: Nit:
```suggestion
   * @protected
```

dpopp07: Good catch, thanks!

PR #89, opened 2020-04-28
Introduce a token-manager class for token handling

vmatyusGitHub:
The new `TokenManager` class will be responsible only for the token handling specifically.
Changed some private fields to protected: `tokenInfo`, `expireTime`, `refreshTime`.
Placed `getCurrentTime` into the utils module to be globally usable.

##### Checklist

- [x] tests are included

### DIFF
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -29,6 +29,8 @@ export type JwtTokenManagerOptions = TokenManagerOptions;
  * to retrieve the bearer token from intended sources.
  */
 export class JwtTokenManager extends TokenManager {
+  protected tokenName: string;
+  protected tokenInfo: any;
 
   /**
    * Create a new [[JwtTokenManager]] instance.
@@ -47,6 +49,7 @@ export class JwtTokenManager extends TokenManager {
     super(options);
 
     this.tokenName = 'access_token';
+    this.tokenInfo = {};
   }
 
   /**
@@ -70,9 +73,9 @@ export class JwtTokenManager extends TokenManager {
    */
   protected saveTokenInfo(tokenResponse): void {
     const responseBody = tokenResponse.result || {};
-    const accessToken = responseBody[this.tokenName];
+    this.accessToken = responseBody[this.tokenName];
 
-    if (!accessToken) {
+    if (!this.accessToken) {
       const err = 'Access token not present in response';
       logger.error(err);
       throw new Error(err);
@@ -80,7 +83,7 @@ export class JwtTokenManager extends TokenManager {
 
     // the time of expiration is found by decoding the JWT access token
     // exp is the time of expire and iat is the time of token retrieval
-    const decodedResponse = jwt.decode(accessToken);
+    const decodedResponse = jwt.decode(this.accessToken);
     if (!decodedResponse) {
       const err = 'Access token recieved is not a valid JWT';
       logger.error(err);
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -17,7 +17,7 @@
 import extend = require('extend');
 import jwt = require('jsonwebtoken');
 import logger from '../../lib/logger';
-import {TokenManager, TokenManagerOptions} from "./token-manager";
+import { TokenManager, TokenManagerOptions } from "./token-manager";
 
 /** Configuration options for JWT token retrieval. */
 export type JwtTokenManagerOptions = TokenManagerOptions;
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -15,29 +15,12 @@
  */
 
 import extend = require('extend');
-import { OutgoingHttpHeaders } from 'http';
 import jwt = require('jsonwebtoken');
 import logger from '../../lib/logger';
-import { RequestWrapper } from '../../lib/request-wrapper';
-
-function getCurrentTime(): number {
-  return Math.floor(Date.now() / 1000);
-}
+import {TokenManager, TokenManagerOptions} from "./token-manager";
 
 /** Configuration options for JWT token retrieval. */
-export type TokenManagerOptions = {
-  /** The endpoint for token requests. */
-  url?: string;
-  /** Headers to be sent with every service token request. */
-  headers?: OutgoingHttpHeaders;
-  /**
-   * A flag that indicates whether verification of
-   *   the server's SSL certificate should be disabled or not.
-   */
-  disableSslVerification?: boolean;
-  /** Allow additional request config parameters */
-  [propName: string]: any;
-}
+export type JwtTokenManagerOptions = TokenManagerOptions;
 
 /**
  * A class for shared functionality for parsing, storing, and requesting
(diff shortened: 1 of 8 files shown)


auth/token-managers/token-manager.ts
```diff
@@ -0,0 +1,240 @@
...
+ * request management. Child classes should implement `requestToken()`
+ * to retrieve the bearer token from intended sources.
+ */
+export class TokenManager {
+  protected url: string;
+  protected tokenName: string;
+  protected disableSslVerification: boolean;
+  protected headers: OutgoingHttpHeaders;
+  protected requestWrapperInstance: RequestWrapper;
+  protected tokenInfo: any;
```
mkistler, line 49: While we're abstracting, I think we should delegate `tokenInfo` and `tokenName` to subclasses of `TokenManager`, with only `accessToken` visible at this level.

auth/token-managers/token-manager.ts
```diff
@@ -0,0 +1,240 @@
...
+   */
+  protected requestToken(): Promise<any> {
+    const errMsg = '`requestToken` MUST be overridden by a subclass of JwtTokenManagerV1.';
+    const err = new Error(errMsg);
+    logger.error(errMsg);
+    return Promise.reject(err);
+  }
+
+  /**
+   * Save the token from the response and the calculated expiration time to the object's state.
```
mkistler, line 191: We should be explicit.  This method must set `accessToken` (see earlier comment), `expireTime`, and `refreshTime` based on the `tokenResponse`.

auth/token-managers/token-manager.ts
vmatyusGitHub: Sorry, but I am not catching your point. `TokenInfo` currently in use to check that a token already is stored in the object. Here is a reference to a line where there properties are used:
https://github.com/vmatyusGitHub/node-sdk-core/blob/2269660441b93d426932ec9b0dfb51ac1dc0159a/auth/token-managers/token-manager.ts#L98
In the constructor `TokenInfo` is initialised to an empty object and `tokenName` shall be initialised in a subclass.
Are you suggesting to have only `accesToken` property with type string, this way I shall remove `tokenInfo` and `tokenName` from the object's property list?

auth/token-managers/token-manager.ts
```diff
@@ -0,0 +1,240 @@
...
+  disableSslVerification?: boolean;
+  /** Allow additional request config parameters */
+  [propName: string]: any;
+}
+
+/**
+ * A class for shared functionality for parsing, storing, and requesting
+ * JWT tokens. Intended to be used as a parent to be extended for token
+ * request management. Child classes should implement `requestToken()`
+ * to retrieve the bearer token from intended sources.
```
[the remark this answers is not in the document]
vmatyusGitHub: Fixed in e92c3de

auth/token-managers/token-manager.ts
```diff
@@ -0,0 +1,240 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { OutgoingHttpHeaders } from 'http';
+import logger from '../../lib/logger';
+import { RequestWrapper } from '../../lib/request-wrapper';
+import {getCurrentTime} from "../utils";
+
+/** Configuration options for JWT token retrieval. */
```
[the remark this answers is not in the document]
vmatyusGitHub: Fixed in e92c3de

auth/token-managers/token-manager.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }
+
+  /**
+   * Request a token using an API endpoint.
+   *
+   * @returns {Promise}
+   */
+  protected requestToken(): Promise<any> {
+    const errMsg = '`requestToken` MUST be overridden by a subclass of JwtTokenManagerV1.';
```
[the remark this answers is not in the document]
vmatyusGitHub: Fixed in e92c3de

auth/token-managers/token-manager.ts
vmatyusGitHub: Half fixed in e92c3de

auth/token-managers/token-manager.ts
dpopp07: Since `tokenInfo` is used in the logic contained in this file, I'm fine with it existing here.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -15,47 +15,20 @@
...
 import extend = require('extend');
-import { OutgoingHttpHeaders } from 'http';
 import jwt = require('jsonwebtoken');
 import logger from '../../lib/logger';
-import { RequestWrapper } from '../../lib/request-wrapper';
-
-function getCurrentTime(): number {
-  return Math.floor(Date.now() / 1000);
-}
+import {TokenManager, TokenManagerOptions} from "./token-manager";
```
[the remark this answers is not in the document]
vmatyusGitHub: Fixed in fc4e9b6

auth/token-managers/token-manager.ts
```diff
@@ -0,0 +1,244 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { OutgoingHttpHeaders } from 'http';
+import logger from '../../lib/logger';
+import { RequestWrapper } from '../../lib/request-wrapper';
+import {getCurrentTime} from "../utils";
```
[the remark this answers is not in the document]
vmatyusGitHub: Fixed in fc4e9b6

auth/token-managers/token-manager.ts
mkistler: Here was my thinking: This class requires its subclasses to implement `saveTokenInfo`, which is passed the response of the token request.  `saveTokenInfo` has to set some member variables ... specifically `expireTime`, `refreshTime`, and `tokenInfo`.

But what is `tokenInfo`?  It is just an object that must contain a field named `tokenName` that contains the access token.  From the point of view of a generic `TokenManager`, these are extraneous details.  All it needs is the `accessToken`.  So why not simplify the `TokenManager` class (i.e. make it more general), and simply use `accessToken` set by `saveTokenInfo`, and push the complexity of varying token property names into the subclass, `JWTTokenManager`.

I can live with the code the way it is, but I think it would be better if `tokenInfo` and `tokenName` was pushed down into `JWTTokenManager` and `TokenManager` only dealt with `accessToken`.

dpopp07:
I second @mkistler - I think the suggestion is good but I can live with the current code. @vmatyusGitHub what do you think?

auth/token-managers/token-manager.ts
vmatyusGitHub: Now, I understand your request. Thank you for the more detailed description. I created a modification that would use `accessToken`. In this commit the change can be reviewed: 2f5fe588dbe3d462e0ab98e323ffdc797f955f4f

vmatyusGitHub:
In my custom authentication case also only 1 access token is in use, so I also think it is a valid request. I am open to implement it.

auth/token-managers/token-manager.ts
vmatyusGitHub: I was just wondering why it is important to use `TokenInfo` inside `JwtTokenManager`? Will there be a feature in the future that will be built upon the content of this object?

dpopp07: Are you referring to the _type_ `TokenInfo`? I don't know that we have a specific feature in mind for the future so it is not truly important. However, I think it is worth leaving in as at the very least it provides a developer with knowledge of what they can expect the response to look like.

vmatyusGitHub:
@mkistler and @dpopp07 Thanks for answering my concerns and reviewing my changes.

vmatyusGitHub:
Can I ask any of you to merge this PR?

PR #90, opened 2020-04-29
fix: flag required params as missing when explicitly undefined

dpopp07:
Currently, we are checking for "truthyness" when evaluating if required parameters have been set or not. This is flawed - when required params are set to `false`, they are incorrectly being flagged as missing.

This PR fixes that bug by explicitly checking for the `undefined` value. Test case added to support the fix.

### DIFF
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -113,8 +113,8 @@ export function getMissingParams(
     missing = requires;
   } else {
     missing = [];
-    requires.forEach((require) => {
-      if (!params[require]) {
+    requires.forEach(require => {
+      if (isMissing(params[require])) {
         missing.push(require);
       }
     });
@@ -122,6 +122,17 @@ export function getMissingParams(
   return missing.length > 0
     ? new Error('Missing required parameters: ' + missing.join(', '))
     : null;
+}
+
+/**
+ * Returns true if value is determined to be "missing". Currently defining "missing"
+ * as `undefined`, `null`, or the empty string.
+ *
+ * @param value - the parameter value
+ * @returns boolean
+ */
+function isMissing(value: any): boolean {
+  return value === undefined || value === null || value === '';
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/helper.ts
```diff
@@ -113,8 +113,8 @@ export function getMissingParams(
     missing = requires;
   } else {
     missing = [];
-    requires.forEach((require) => {
-      if (!params[require]) {
+    requires.forEach(require => {
+      if (params[require] === undefined) {
```
mkistler, line 117: According to [this site](https://guide.freecodecamp.org/javascript/falsy-values/):
> There are only six falsey values in JavaScript: `undefined` ,` null` ,`NaN` , `0` , `""` (empty string), and `false` of course.

I can see how we want to allow `false` and disallow `undefined`. Likewise `0` and `NAN` seems like they should be fine.

But I'm concerned about ` null` and `""` (empty string).  I think either of these, if passed for a path parameter, would produce an invalid path.  Most of the other languages have a slightly different check for path parameters that also disallows an empty string.  I'm also concerned about what `null` would yield if passed for a query parameter.

dpopp07: My thinking was that explicit `null` and empty string values should be allowed in the theoretical sense but I agree that those values probably wouldn't be very meaningful in actual practice. Do you think we should go ahead and check for that here, or issue a warning in the logger, or add a separate check for specific parameter types?

mkistler: I think what I am most concerned about are cases where a "bizarre" error can occur, like an empty string path parameter.  In this case, the request will probably be a `404` or even some kind of "maformed url", which users will find confusing.  I'm less concerned about an empty string passed in a request body that actually makes it to the server and then the server rejects with a (hopefully) meaningful error code.

Would it be a problem to reject `undefined` or `null` or `""` -- that is, continue to treat those values as not supplied?  That would fix the problem for `false`, `0`, and `NAN`, but not overstep into values that might be problematic.

dpopp07: I'm fine with that. If it ends up being a problem, we can always change it again. I'll make that change

PR #95, opened 2020-05-13
fix: Remove trailing slash from url for extra cases

rmkeezer:
Covered extra cases where trailing slash was not being removed when making a request.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/base-service.ts
+++ b/lib/base-service.ts
@@ -141,7 +141,9 @@ export class BaseService {
    * @param {string} url The base URL for the service.
    */
   public setServiceUrl(url: string): void {
-    this.baseOptions.serviceUrl = url;
+    if (url) {
+      this.baseOptions.serviceUrl = stripTrailingSlash(url);
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


lib/base-service.ts
```diff
@@ -141,7 +141,9 @@ export class BaseService {
    * @param {string} url The base URL for the service.
    */
   public setServiceUrl(url: string): void {
-    this.baseOptions.serviceUrl = url;
+    if (url) {
```
MasterOdin, line 144: Given a parameter type of just `string`, this if statement would be superfluous. Should the type be changed to `url?: string` or the if statement removed?

dpopp07: Yeah I think that's a fair point, the `if` statement probably doesn't need to be there. Or should be stricter and ensure that it's actually a string. In the JS case (which we think is most of our users), the type or requiredness isn't actually enforced. And `stripTrailingSlash` uses the `.replace()` method so it will throw a TypeError if `url` isn't a string.

I agree that the `if` statement as is isn't really that useful

PR #104, opened 2020-08-05
fix: support `disableSslVerification` when user provides custom httpsAgent

dpopp07:
Before, if a user set `disableSslVerification` to true and also provided their own httpsAgent (to configure a proxy, etc), the `disableSslVerification` handling would be overwritten. We would create an httpsAgent with the required property, but the user's agent would get used instead.

Now, if the user provides their own agent, the code will update the user's agent to add the required property. This way, if a user ever specifies `disableSslVerification` to true, it will always be handled properly at the agent level.

### DIFF
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -15,6 +15,7 @@
  */
 
 import axios from 'axios';
+import { AxiosRequestConfig } from 'axios';
 import axiosCookieJarSupport from 'axios-cookiejar-support';
 import FormData = require('form-data');
 import https = require('https');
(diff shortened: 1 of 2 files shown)


lib/request-wrapper.ts
```diff
@@ -26,9 +27,6 @@ import logger from './logger';
 const isBrowser = typeof window === 'object';
 const globalTransactionId = 'x-global-transaction-id';
 
-// Limit the type of axios configs to be customizable
-const allowedAxiosConfig = ['transformRequest', 'transformResponse', 'paramsSerializer', 'paramsSerializer', 'timeout', 'withCredentials', 'adapter', 'responseType', 'responseEncoding', 'xsrfCookieName', 'xsrfHeaderName', 'onUploadProgress', 'onDownloadProgress', 'maxContentLength', 'validateStatus', 'maxRedirects', 'socketPath', 'httpAgent', 'httpsAgent', 'proxy', 'cancelToken', 'jar'];
-
```
dpopp07, line 31: I never liked this code and it was actually causing some strange behavior, so I removed it entirely.

PR #109, opened 2020-09-17
feat(IAM Authenticator): add support for optional 'scope' property

jorge-ibm:
The IAM /identity/token (get token) operation supports an optional "scope" form parameter which can be used to obtain an IAM access token having a limited set of scopes in this it can be used.

This commit adds support for the "Scope" field within the IamTokenManager & IamAuthenticator class. If set, the value of this field is sent as the "scope" form param in the "get token" request body when obtaining an access token.

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2167

##### Checklist

- [x] tests are included

### DIFF
--- a/auth/token-managers/iam-token-manager.ts
+++ b/auth/token-managers/iam-token-manager.ts
@@ -35,12 +35,14 @@ function onlyOne(a: any, b: any): boolean {
 }
 
 const CLIENT_ID_SECRET_WARNING = 'Warning: Client ID and Secret must BOTH be given, or the header will not be included.';
+const SCOPE = 'scope';
 
 /** Configuration options for IAM token retrieval. */
 interface Options extends JwtTokenManagerOptions {
   apikey: string;
   clientId?: string;
   clientSecret?: string;
+  scope?: string;
 }
 
 /**
@@ -53,6 +55,7 @@ export class IamTokenManager extends JwtTokenManager {
   private apikey: string;
   private clientId: string;
   private clientSecret: string;
+  private scope: string;
 
   /**
    *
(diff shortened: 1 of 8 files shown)


auth/token-managers/iam-token-manager.ts
```diff
@@ -35,12 +35,14 @@ function onlyOne(a: any, b: any): boolean {
...
 
 const CLIENT_ID_SECRET_WARNING = 'Warning: Client ID and Secret must BOTH be given, or the header will not be included.';
+const SCOPE = 'scope';
 
 /** Configuration options for IAM token retrieval. */
 interface Options extends JwtTokenManagerOptions {
   apikey: string;
   clientId?: string;
   clientSecret?: string;
+  string?: string;
```
dpopp07, line 45: I think this is meant to be `scope`?
```suggestion
  scope?: string;
```

auth/token-managers/iam-token-manager.ts
```diff
@@ -143,6 +161,10 @@ export class IamTokenManager extends JwtTokenManager {
       }
     };
 
+    if (this.scope && this.scope !== '') {
```
dpopp07, line 164: The second check is redundant - if it is an empty string `this.scope` will return `false`
```suggestion
    if (this.scope) {
```

auth/token-managers/iam-token-manager.ts
```diff
@@ -35,12 +35,14 @@ function onlyOne(a: any, b: any): boolean {
 }
 
 const CLIENT_ID_SECRET_WARNING = 'Warning: Client ID and Secret must BOTH be given, or the header will not be included.';
+const SCOPE = 'scope';
```
dpopp07, line 38: I don't know that a constant here is really necessary but I also don't have a problem with it. Up to you but I thought I would at least say that. If you find it helpful, I'm fine with it

test/unit/get-authenticator-from-environment.test.js
```diff
@@ -89,6 +89,13 @@ describe('Get Authenticator From Environment Module', () => {
       getAuthenticatorFromEnvironment(SERVICE_NAME);
     }).toThrow();
   });
+
+  it('should get iam authenticator and set the scope', () => {
+    setUpIamPayloadWithScope();
+    const authenticator = getAuthenticatorFromEnvironment(SERVICE_NAME);
+    expect(authenticator).toBeInstanceOf(IamAuthenticator);
+    expect(authenticator.scope).toBe('john snow');
```
dpopp07, line 97: jon* snow 🙂

dpopp07 (approved):
Awesome work. There is just one small typo that **must be fixed** before merging but I am approving to avoid the need for a re-review. I left some other suggestions that you can take or leave at your own discretion

auth/token-managers/iam-token-manager.ts
jorge-ibm: oh wow, yep completely missed this..

test/unit/get-authenticator-from-environment.test.js
jorge-ibm: wow, I always thought it was John 😞

padamstx (approved):
LGTM

PR #110, opened 2020-09-22
feat: upgrade axios and re-enable gzipping responses

dpopp07:
I spent some time today reading through the `axios` docs and issue and I _think_ it should be fine to upgrade to the latest version (0.20.0). It seems to be much more actively maintained than it has been the past year or so, which makes me more comfortable about upgrading. There seem to have been [a couple of unintentional breaking changes](https://github.com/axios/axios/issues/3294#issuecomment-696601393) in this version but neither of them affect us because of how we're using the library in our code.

In addition to this version bump, I added the `gzip` encoding on responses back in, which was waiting on a fix within `axios`.

### DIFF
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -202,11 +202,8 @@ export class RequestWrapper {
       headers = extend(true, {}, headers, multipartForm.getHeaders());
     }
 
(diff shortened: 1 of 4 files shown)


padamstx (approved):
LGTM

jorge-ibm (approved):
LGTM

PR #111, opened 2020-10-02
feat: add support for compressing request bodies

dpopp07:
A flag can now be passed to the BaseService that will enable gzip compression on the request body data. This flag is passed down to the RequestWrapper class where the compression code lives. Note that no code needs to be changed in the BaseService - the options passed into its constructor will be carried through to the RequestWrapper. In the generated code, I am planning on attaching the flag to the options passed into the BaseService constructor.

If there is a problem during compression, the code moves alone with the request, just uncompressed. If this is not the behavior we want, I can change it.

### DIFF
--- a/lib/base-service.ts
+++ b/lib/base-service.ts
@@ -146,6 +146,15 @@ export class BaseService {
   }
 
   /**
+   * Turn request body compression on or off.
+   *
+   * @param {boolean} setting Will turn it on if 'true', off if 'false'. 
+   */
+  public setEnableGzipCompression(setting: boolean): void {
+    this.requestWrapperInstance.compressRequestData = setting;
+  }
+
+  /**
    * Configure the service using external configuration
    *
    * @param {string} serviceName The name of the service. Will be used to read from external
@@ -205,14 +214,21 @@ export class BaseService {
     const properties = readExternalSources(serviceName);
 
     if (properties !== null) {
-      // the user can define two client-level variables in the credentials file: url and disableSsl
-      const { url, disableSsl } = properties;
+      // the user can define the following client-level variables in the credentials file:
+      // - url
+      // - disableSsl
+      // - enableGzip
+
+      const { url, disableSsl, enableGzip } = properties;
 
       if (url) {
         results.serviceUrl = stripTrailingSlash(url);
       }
       if (disableSsl === true) {
         results.disableSslVerification = disableSsl;
+      }
+      if (enableGzip === true) {
+        results.enableGzipCompression = enableGzip;
       }
     }
 
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -19,20 +19,25 @@ import { AxiosRequestConfig } from 'axios';
 import axiosCookieJarSupport from 'axios-cookiejar-support';
 import extend = require('extend');
 import FormData = require('form-data');
+import { OutgoingHttpHeaders } from 'http';
 import https = require('https');
+import isStream = require('isstream');
 import querystring = require('querystring');
-import { PassThrough as readableStream } from 'stream';
+import zlib = require('zlib');
 import { buildRequestFileObject, getMissingParams, isEmptyObject, isFileData, isFileWithMetadata, stripTrailingSlash } from './helper';
 import logger from './logger';
+import { streamToPromise } from './stream-to-promise';
 
 const isBrowser = typeof window === 'object';
 const globalTransactionId = 'x-global-transaction-id';
 
 export class RequestWrapper {
   private axiosInstance;
+  private compressRequestData: boolean;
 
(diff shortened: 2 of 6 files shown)


lib/request-wrapper.ts
```diff
@@ -311,6 +320,29 @@ export class RequestWrapper {
 
     return error;
   }
+
+  private gzipRequestBody(data: Object, headers: OutgoingHttpHeaders): Buffer|Object {
```
padamstx, line 324: If gzip compression is enabled, we should compress the request body regardless of the type of data that is present.  i.e. this would apply to more than just JSON objects that need to be stringified, such as a text/plain or application/octet stream type request body.

lib/request-wrapper.ts
```diff
@@ -311,6 +320,29 @@ export class RequestWrapper {
...
+    // skip compression if user has overridden the encoding header
+    if (headers['content-encoding']) {
```
padamstx, line 326: A couple things with this...
1) Do we not have to worry about the case of header names in node?   i.e. if the user set the "Content-Encoding" header (what is considered the canonicalized form of "content-encoding") would that register here?
2) Content-Encoding is one of those headers that could conceivably have more than one value.   I think the check here should be whether or not the header already has the value "gzip" rather than whether or not Content-Encoding is already set or not.

padamstx (changes requested):
There are a couple things that need to be addressed, plus we should also support enabling gzip compression via an external config property... e.g. `MYSERVICE_ENABLE_GZIP=true`.

lib/request-wrapper.ts
dpopp07: 1. I think we do have to worry about the case. I was going purely off of the design wiki but I see that the other cores use `Content-Encoding` so I'll change it.
2. That sounds good 👍

dpopp07:
All feedback should now be addressed

lib/base-service.ts
```diff
@@ -205,15 +214,22 @@ export class BaseService {
...
+
+      const { url, disableSsl, enableGzip } = properties;
 
       if (url) {
         results.serviceUrl = stripTrailingSlash(url);
       }
       if (disableSsl === true) {
         results.disableSslVerification = disableSsl;
       }
+      if (enableGzip === true) {
```
padamstx, line 230: If the `<service>_ENABLE_GZIP` property is defined with any value at all, then we want to use the value whether it is "true" or "false".    This behavior will allow for a default value being set in the service instance by the generated service ctor, and will allow the user to override that default value (in either direction) by setting the `<service>_ENABLE_GZIP` config property to the opposite value.

test/unit/request-wrapper.test.js
```diff
@@ -680,3 +718,156 @@ describe('formatError', () => {
...
+    expect(data).toBe(null);
+  });
+
+  it('should compress json data into a buffer', async () => {
+    let data = { key: 'value' };
+    const headers = {};
+
+    const jsonSpy = jest.spyOn(JSON, 'stringify');
+
+    data = await requestWrapperInstance.gzipRequestBody(data, headers);
```
padamstx, line 764: We also should verify that Content-Encoding was set to "gzip".  There are a few other tests where this check should be added as well.

padamstx (changes requested):
Unless I mis-read something, I think there's an issue with the handling of the `<service>_ENABLE_GZIP` config property.   And I had one other comment re: checking the Content-Encoding header as part of the successful tests.

test/unit/request-wrapper.test.js
dpopp07: I am checking that explicitly in its own test [here](https://github.com/IBM/node-sdk-core/pull/111/files#diff-5857e7caf9c146751fb8caef9bae6024R813) but I'm happy to add that check to the other tests as well if you think I should.

lib/base-service.ts
dpopp07: The way I'm handling this is by storing a Boolean value in the request wrapper instance stored in the base service, because that's the class that actually uses the value. When the instance is created, the stored variable is [set to `true` if explicitly set and `false` if not](https://github.com/IBM/node-sdk-core/pull/111/files#diff-3cf5813ed1f8d97bdaad8c743e775fd8R40). When the user overrides the variable with the setter, the value in the request wrapper is updated.

I think this behavior exactly matches what you are describing, I'm just keeping the value buried in the request wrapper (where it will be used) rather than storing it in two different places.

test/unit/request-wrapper.test.js
dpopp07: I went ahead and added that check to all gzip tests 👍

jorge-ibm (approved):
Looks good! Looks like Phil covered the concerns I had.

padamstx (approved):
Looks good!

PR #112, opened 2020-10-27
chore: appscan configuration file

christiancompton:
The changes in the `appscan-config` file will exclude the `/scripts` directory in our security scanning tool.


padamstx (approved):
I'll approve now to avoid another review, but we discussed a change needed to the config file.

PR #114, opened 2020-11-19
build: update semantic-release to v17

dpopp07:
This version of `semantic-release` includes breaking changes that I have addressed in the
`.releaserc` file. The only change that affects us is that "branch" was renamed to "branches"
to accept multiple values. It can still accept a single string value so I only renamed the property value.

Opening this to replace #113

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,5 +1,5 @@
 {
-  "branch": "master",
+  "branches": "master",
   "verifyConditions": ["@semantic-release/changelog", "@semantic-release/npm", "@semantic-release/git"],
   "prepare": ["@semantic-release/changelog", "@semantic-release/npm", "@semantic-release/git"],
   "publish": [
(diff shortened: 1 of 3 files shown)


.releaserc
```diff
@@ -1,5 +1,5 @@
 {
-  "branch": "master",
+  "branches": "master",
```
padamstx, line 2: Thanks for tagging me on this.  I need to update the PS node-sdk with this change.   Apparently "master" must be the default branch that it operates on (thankfully) :)

padamstx (approved):
LGTM

.releaserc
dpopp07: Yes, "master" is the default so I actually thought about simply removing this line from the file. I only left it in in case we change the branch system down the line for some reason, I didn't want anyone to be confused when their branch wasn't found 🙂

PR #121, opened 2021-02-10
fix(build): main migration

christiancompton:
This changeset updates the automation and documentation to support the new `main` default branch.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,5 +1,5 @@
 {
-  "branches": "master",
+  "branches": "main",
   "verifyConditions": ["@semantic-release/changelog", "@semantic-release/npm", "@semantic-release/git"],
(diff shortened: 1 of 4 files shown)


padamstx (approved):
Changes look good, but I have one more.  While reviewing the PR, I was curious why no changes were needed to the .travis.yml file and discovered that we have this as part of the "deploy" stage:
```
    repo: IBM/node-sdk-core
```
Please change this line to be:
```
    branch: main
```
Discussed this with @dpopp07 and we both agree on this change.  Not sure why we were using the "repo" setting to begin with :)

Edit:  oh, one more thing... can you please change your commit message to something like:
```
fix(build): main migration
```
the "fix" type will cause a new patch level, which means we'll exercise the semantic release step with the branch name change, etc.

dpopp07 (approved):
Looks good, but I second Phil's suggestion to go ahead use only run deploy builds on the `main` branch

PR #122, opened 2021-02-19
feat(iam-authenticator): expose refresh token with a getter

dpopp07:
The `IamAuthenticator` class now makes the refresh token available to users through
a public method, `getRefreshToken`. This provides support to services that require
the refresh token as a part of their authentication scheme.

Note that this is only supported for the IAM authenticator. Let me know if we need the same support in the CP4D authenticator.


padamstx (approved):
LGTM

PR #127, opened 2021-03-11
fix: persist `enableGzipCompression` setting on the base service

dpopp07:
When a new `BaseService` instance is created, the class settings are stored in an object
called `baseOptions`. Currently, when the `setEnableGzipCompression` method is called,
the `enableGzipCompression` setting is not changed and might no longer reflect the state
of the setting (because it is primarily tracked in the underlying `RequestWrapper` class).

This PR addresses this by persisting the setting in the `baseOptions` object when that method
is called so that the object always reflects the accurate value for the setting.


padamstx (approved):
LGTM

PR #128, opened 2021-03-12
Add getQueryParam utility method to support pagination

mkistler:
This PR adds a small utility method to extract the value of a query parameter from a URL string. This will be used in the pagination support for Node.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -96,6 +96,21 @@ export function stripTrailingSlash(url: string): string {
 }
 
 /**
+ * Return a query parameter value from a URL
+ *
+ * @param {string} urlStr - the url string.
+ * @param {string} param - the name of the query parameter
+ *                     whose value should be returned
+ * @returns {string} the value of the `param` query parameter
+ * @throws if urlStr is an invalid URL
+ */
+export function getQueryParam(urlStr: string, param: string): string {
+  // The base URL is a dummy value just so we can process relative URLs
+  const url = new URL(urlStr, 'https://foo.bar');
+  return url.searchParams.get(param);
+}
+
+/**
  * Validates that all required params are provided
  * @param params - the method parameters.
  * @param requires - the required parameter names.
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM

lib/helper.ts
```diff
@@ -95,6 +95,20 @@ export function stripTrailingSlash(url: string): string {
...
+ * Return a query parameter value from a URL
+ *
+ * @param {string} urlStr - the url string.
+ * @param {string} param - the name of the query parameter
+ *                     whose value should be returned
+ * @returns {string} the value of the `param` query parameter
+ * @throws if urlStr is an invalid URL
+ */
+export function getQueryParam(urlStr: string, param: string): string {
+  const url = new URL(urlStr, 'https://foo.bar');
```
dpopp07, line 108: Do you mind adding a comment explaining that `'https://foo.bar'` is there to support relative URLs? Maybe this is more common than I realize but I've never seen that pattern and I was a little confused. Upon looking it up it made perfect sense but a comment may be helpful for future people

test/unit/get-query-param.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+  it('should return null when the query string is invalid', function() {
+    const nextUrl = '/api/v1/offerings?start%XXfoo';
+    expect(getQueryParam(nextUrl, 'start')).toBeNull();
+  });
+
+  it('should return the first value when the query string has duplicate parameters', function() {
+    const nextUrl = '/api/v1/offerings?start=foo&start=bar&limit=10';
+    expect(getQueryParam(nextUrl, 'start')).toBe('foo');
+  });
```
dpopp07, line 51: Great tests! 👏

dpopp07 (approved):
Looks good! 👍

lib/helper.ts
mkistler: Good suggestion!  I will add a comment.

PR #129, opened 2021-03-16
fix: enable symbolic-link credentials files

dpopp07:
Currently, credentials files are explicitly required to real files (not symbolic links).
This was to exclude directories, not to exclude symbolic links, so it is a bug.

This PR adds explicit support for symbolic links. This should be a robust solution, as
I do not believe we want to support any of the other types the file could be (sockets,
block devices, etc).


padamstx (approved):
LGTM

PR #132, opened 2021-04-01
fix: stub sdk-test-helpers in browser environments

dpopp07:
The "sdk-test-helpers" file contains helper functions that generated SDKs use
in their generated unit tests. These functions are never actually needed when
running standard operational code, therefore they are never used in the browser.

The `expect` library that we use for testing has a dependency that isn't browser
friendly in certain scenarios. This causes the core and any SDKs importing it
to throw errors in these scenarios (like the React framework).

This change stubs the sdk-test-helpers in browser environments the same way we
stub the credentials file reading code. I've verified that this fix works in the problematic environments

cc @apaparazzi0329

### DIFF
--- a/lib/sdk-test-helpers.browser.ts
+++ b/lib/sdk-test-helpers.browser.ts
@@ -1,0 +1,24 @@
+// Dummy for browser
+export function checkUrlAndMethod(options, url: string, method: any) {
+  return;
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM

PR #134, opened 2021-05-03
build: upgrade dependencies to avoid vuln alerts

padamstx:
This PR upgrades the jest and typedoc dependencies to avoid pulling in vulnerable versions of downstream dependencies.


dpopp07 (approved):
Looks good! 👍

PR #135, opened 2021-05-07
chore: update pkglock.json

christiancompton:
This PR updates the package-lock.json to address a disclosed vulnerability in a transitive dependency.


dpopp07 (approved):
This is difficult to review but it seems fine and the tests are all passing, so it looks okay to me 👍

padamstx:
I took a glance at the changes after seeing @dpopp07 's comment, and I think the file changed so much due to @christiancompton using an older version of npm.
I think Dustin and I are using npm v7 which uses the lockfile v2 if I understand things correctly.

dpopp07:
You may be right, though I moved back to npm v6 a while ago because I was having issues with v7

PR #136, opened 2021-05-07
fix: upgrade package-lock.json

padamstx:
This PR updates package-lock.json back to v2 (npm v7.x).  Otherwise, there shouldn't be any actual changes to dependency versions.   An `npm audit` run in my local sandbox yielded 0 vulnerabilities (knock on wood!).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14637 @@
 {
   "name": "ibm-cloud-sdk-core",
   "version": "2.10.3",
-  "lockfileVersion": 1,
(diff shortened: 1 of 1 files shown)


padamstx:
btw, I used a "fix" commit in order to trigger a new patch level.  I thought that would be good since Christian's previous change fixed some vulnerabilities.

dpopp07 (approved):
Fine by me!

I'm honestly not 100% sure what difference it makes to use one version of the lockfile vs the other but I'm fine to keep things at their latest version

PR #138, opened 2021-05-26
feat: add `BaseService.constructServiceURL` method

ajhynes7:
This method can be used by an API to construct a service URL from a parameterized URL.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/base-service.ts
+++ b/lib/base-service.ts
@@ -61,6 +61,50 @@ export class BaseService {
 
   static DEFAULT_SERVICE_NAME: string;
 
+  /**
+   * Constructs a service URL by formatting a parameterized URL.
+   *
+   * @param {string} parameterizedUrl URL that contains variable placeholders, e.g. '{scheme}://ibm.com'.
+   * @param {Map<string, string>} defaultUrlVariables Map from variable names to default values.
+   *  Each variable in the parameterized URL must have a default value specified in this map.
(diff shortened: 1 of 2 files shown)


lib/base-service.ts
```diff
@@ -61,6 +61,50 @@ export class BaseService {
...
+   */
+  static constructServiceURL(
+    parameterizedUrl: string,
+    defaultUrlVariables: Map<string, string>,
+    providedUrlVariables: Map<string, string> | null
+  ): string {
+    // If null was passed, we set the variables to an empty map.
+    // This results in all default variable values being used.
+    if (providedUrlVariables === null) {
+      providedUrlVariables = new Map<string, string>();
```
dpopp07, line 83: Quick initial question - is there a reason that you used Maps instead of basic objects? e.g. `{}`

ajhynes7: I'm implementing the same method for Java and Go using a map, so I chose the `Map` class for consistency. The class provides some useful methods like `.has(key)`. I also chose the `Map` class because the variables are intended to just have keys and values - nothing more nested than that.

dpopp07: That all makes sense! I asked because I don't see them used very often but they seem great for this use case 👍

padamstx (approved):
LGTM!

dpopp07 (approved):
Looks good! 👍

PR #140, opened 2021-06-03
test: add tests that explicitly set the Host request header

padamstx:
This PR adds a couple of request-wrapper unit tests to verify that when the Host header is explicitly set in the Axios options, it appears to be included in the http request mesage.


dpopp07 (approved):
Looks good!

PR #141, opened 2021-06-04
fix: change constructServiceURL from method to function

ajhynes7:
Now it can be used as a utility function instead of an inherited method.

An SDK class can now include a `constructServiceURL` method with a different signature than this function.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -253,7 +253,7 @@ export function toLowerKeys(obj: Object): Object {
  *  the default variable value will be used instead.
  * @returns {string} The formatted URL with all variable placeholders replaced by values.
  */
-export function constructServiceURL(
+export function constructServiceUrl(
   parameterizedUrl: string,
   defaultUrlVariables: Map<string, string>,
   providedUrlVariables: Map<string, string> | null
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -241,3 +241,47 @@ export function toLowerKeys(obj: Object): Object {
   }
   return lowerCaseObj;
 }
+
+/**
+ * Constructs a service URL by formatting a parameterized URL.
+ *
+ * @param {string} parameterizedUrl URL that contains variable placeholders, e.g. '{scheme}://ibm.com'.
+ * @param {Map<string, string>} defaultUrlVariables Map from variable names to default values.
+ *  Each variable in the parameterized URL must have a default value specified in this map.
+ * @param {Map<string, string>} providedUrlVariables Map from variable names to desired values.
+ *  If a variable is not provided in this map,
+ *  the default variable value will be used instead.
+ * @returns {string} The formatted URL with all variable placeholders replaced by values.
+ */
+export function constructServiceURL(
+  parameterizedUrl: string,
+  defaultUrlVariables: Map<string, string>,
+  providedUrlVariables: Map<string, string> | null
+): string {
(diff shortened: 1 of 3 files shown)


lib/helper.ts
```diff
@@ -241,3 +241,47 @@ export function toLowerKeys(obj: Object): Object {
...
+ *
+ * @param {string} parameterizedUrl URL that contains variable placeholders, e.g. '{scheme}://ibm.com'.
+ * @param {Map<string, string>} defaultUrlVariables Map from variable names to default values.
+ *  Each variable in the parameterized URL must have a default value specified in this map.
+ * @param {Map<string, string>} providedUrlVariables Map from variable names to desired values.
+ *  If a variable is not provided in this map,
+ *  the default variable value will be used instead.
+ * @returns {string} The formatted URL with all variable placeholders replaced by values.
+ */
+export function constructServiceURL(
```
dpopp07, line 256: I only just now thought of this, but since we're changing it already... we don't fully capitalize "URL" elsewhere in the core. For example, the BaseService class has a method called `setServiceUrl`. I think we probably want to be consistent and call this `constructServiceUrl`. Sorry for not catching that earlier!

test/unit/parameterized-url.test.js
```diff
@@ -1,4 +1,4 @@
-const { BaseService } = require('../../dist/lib/base-service');
+const { constructServiceURL } = require('../../dist/lib/helper');
```
padamstx, line 1: Question: will moving this function to helpers.ts make it more difficult to import it in the generated SDK code?
@dpopp07 any idea?

lib/helper.ts
padamstx: I noticed this too, and tbh we have the same discrepancy in the java core.  I only noticed that AFTER the PR was merged in :)
Let's make this change here while we're changing things anyway.

padamstx: BTW, I'm not suggesting we change the Java core :)

test/unit/parameterized-url.test.js
dpopp07: @padamstx Good question - no, it won't. All of the helper methods are exported at the top level of the core so the import will be simple:

```node
const { constructServiceUrl, <any-other-imports> } = require('ibm-cloud-sdk-core');
```

dpopp07 (approved):
Looks good! 👍

test/unit/parameterized-url.test.js
padamstx: ok, sounds good, but I assume this means that when we generate the service code for a service, we'll need to be careful about correctly importing `constructServiceUrl` from the core when we will be generating a service-specific `constructServiceUrl()` function that will reference the function in the core.

dpopp07: Oh, because of name conflicts? That could be a legitimate issue, though it should be easy to get around in a compatible way

padamstx: I wasn't thinking about name conflicts actually.  But in the node generator, when we process an API definition, we'd generate the constructServiceUrl() function in the service .ts file ONLY if we detect server variables in one of the "servers" entries.
So only in that case would we want to import the "constructServiceUrl" function from the core, since that's the only scenario where we'd use it.

dpopp07: Oh, I see. That makes sense

PR #145, opened 2021-06-10
fix: remove default limit on request body length

dpopp07:
The `axios` library sets an arbitrary maximum length for request bodies and
throws an error if this length is exceeded. This has caused some confusion for
users and none of the other core libraries enforce such a limit. Services are
expected to handle request body length themselves. So this commit overrides the
default to be "Infinity", just like we do for response body limits. Users still
have the option to override this limit if they wish.


apaparazzi0329:
LGTM

padamstx (approved):
LGTM

PR #147, opened 2021-06-14
fix(build): revert inadvertently-committed directories

padamstx:
This PR contains the following changes:
- it reverts a recent commit that inadvertently committed the coverage, dist, and node_modules directories
- temporarily removes the generation and publishing of jsdocs until we can figure out the correct course of action
- bumps the version of normalize-url package to avoid a high-severity vulnerability

PS: Wow, my PR is removing 3.3M lines :)


dpopp07 (approved):
Looks good, thanks for jumping on this so quickly

PR #149, opened 2021-07-23
refactor: remove specific handling for "watson-token" query params

dpopp07:
We currently have a special hack to handle watson tokens in a unique way.
This used to be required for Watson services but they have not used this
method of authentication in years. This is supposed to be a service-agnostic
package, so this commit refactors the code to avoid the specific handling.

I also edited a jsdoc file (even though I don't think it is used) and added
a test for this small module.

### DIFF
--- a/lib/querystring.ts
+++ b/lib/querystring.ts
@@ -1,29 +1,14 @@
-/* eslint-disable prettier/prettier, arrow-body-style */
-
 /**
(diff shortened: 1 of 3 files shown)


padamstx (approved):
LGTM

barrett-schonefeld (approved):
Changes and test look good!

PR #150, opened 2021-08-03
refactor: move iam-specific functionality to a shared class 

dpopp07:
This will allow other token managers that interact with the IAM service to share configuration and functionality. This change is being made in advance of a new container-based token manager being added that will utilize this shared functionality, as it will interact with the IAM service as well.

cc @pyrooka

### DIFF
--- a/test/unit/iam-token-manager.test.js
+++ b/test/unit/iam-token-manager.test.js
@@ -29,10 +29,7 @@ const IAM_RESPONSE = {
   },
 };
 
-const CLIENT_ID_SECRET_WARNING =
-  'Warning: Client ID and Secret must BOTH be given, or the header will not be included.';
-
-describe('iam_token_manager_v1', () => {
+describe('IAM Token Manager', () => {
   beforeEach(() => {
     mockSendRequest.mockReset();
   });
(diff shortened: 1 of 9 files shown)


test/unit/iam-token-manager.test.js
```diff
@@ -119,294 +141,5 @@ describe('iam_token_manager_v1', () => {
     const token = await instance.getToken();
     expect(token).toBe(ACCESS_TOKEN);
     expect(requestMock).not.toHaveBeenCalled();
-    done();
-  });
-
```
dpopp07, line 124: All of these tests below were consolidated and are covered by the tests for the new IAM request based token manager

test/unit/logger.test.js
```diff
@@ -17,11 +17,11 @@ describe('Logger', () => {
...
-    expect(logger.warn.enabled).toBe(false);
-    expect(logger.verbose.enabled).toBe(false);
-    expect(logger.info.enabled).toBe(false);
-    expect(logger.error.enabled).toBe(false);
-    expect(logger.debug.enabled).toBe(false);
+    expect(logger.warn.enabled).toBeFalsy();
+    expect(logger.verbose.enabled).toBeFalsy();
+    expect(logger.info.enabled).toBeFalsy();
+    expect(logger.error.enabled).toBeFalsy();
+    expect(logger.debug.enabled).toBeFalsy();
```
dpopp07, line 24: These were failing locally. Honestly, I'm not sure how they were ever passing. We never set those parameters to `false`

padamstx (approved):
LGTM

PR #152, opened 2021-08-11
fix: use typescript 3.8.x to build the published code

dpopp07:
Typescript v3.9 contains breaking changes relative to v3.8. These changes
affect generated SDK projects that use a core package built with TS 3.9.
Specifically, the generated unit tests in the SDK project are prevented
from running. In the future, we may want to upgrade to Typescript v4 - this
will likely require updates to our generated code to be compatible. Until
then, we need to stick with 3.8.x.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,26 +13,26 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.14.9",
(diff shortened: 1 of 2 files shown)


barrett-schonefeld (approved):
👍

padamstx (approved):
LGTM

PR #155, opened 2021-08-17
feat: add method `getHttpClient` to get underlying axios instance of base service

dpopp07:
This new method is added to the base service class so that all initialized SDK client objects
have access to the underlying axios instance used to make requests. This instance is currently
hidden as a private class member. This functionality is already supported in the other language
cores, so this brings feature parity


barrett-schonefeld (approved):
Looks good!

padamstx (approved):
:+1:

PR #156, opened 2021-08-23
feat: add enableRetries and disableRetries

barrett-schonefeld:
Hey, folks, I still need to add tests and documentation, but I want to show the general approach.

I tried fixing some type-related issues as well. I successfully resolved a couple of type-related issues (not bugs but improvements to type safety). However, some fixes were not problems at compile time, but I was getting odd runtime errors. 

For example, you will see that I made `compressRequestData` public. I initially added `setCompressRequestData()` to the `RequestWrapper` class, but I was getting a `TypeError: this.requestWrapperInstance.setCompressRequestData is not a function` error. The TypeScript compiler had no issues with this, VSCode jumped to the function definition no problem, but the tests did not like the change. For now, I just made the data public to avoid spending too much time on something potentially insignificant, but do feel free to comment improved solutions on the type-related changes.

##### Checklist

- [ ] tests are included


lib/request-wrapper.ts
```diff
@@ -240,6 +260,7 @@ export class RequestWrapper {
       headers,
       params: qs,
       data,
+      raxConfig: this.retryConfig ?? this.retryConfig,
```
padamstx, line 263: I went ahead and looked up the ?? operator :) and I learned that it is called the "nullish coalescing operator" (can't make this up!), but I'm wondering what good does it do to put the same thing on both sides of the ?? operator?  It looks like you'll use `this.retryConfig` if it is defined and non-null and `this.retryConfig` if it null or undefined :)   At least according to my understanding of this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

lib/base-service.ts
```diff
@@ -170,6 +170,22 @@ export class BaseService {
     return this.requestWrapperInstance.getHttpClient();
   }
 
+  /**
+   * Enable retries for unfulfilled requests.
+   *
+   * @param {Map<string, number>} retryConfig configuration for retries
```
padamstx, line 176: We'll need to add some info about the supported keys/fields within the map.   But really... isn't retryConfig just an object?

lib/request-wrapper.ts
```diff
@@ -36,10 +37,29 @@ import {
...
+}
+
 export class RequestWrapper {
-  private axiosInstance;
+  private axiosInstance: AxiosInstance;
 
-  private compressRequestData: boolean;
+  private retryInterceptorId: number;
+
+  private retryConfig: rax.RetryConfig;
```
padamstx, line 60: I wonder if we should rename this field to be "raxConfig" to make it really clear that it is the config used with retry-axios (rax).

lib/request-wrapper.ts
```diff
@@ -346,6 +368,40 @@ export class RequestWrapper {
     return this.axiosInstance;
   }
 
+  private static getRetryConfig(
+    axiosInstance: AxiosInstance,
+    retryConfig?: RetryOptions
```
padamstx, line 373: suggestion: rename this param to be "retryOptions" to make it clear that it is the set of options that we're exposing to our users

lib/request-wrapper.ts
```diff
@@ -346,6 +368,40 @@ export class RequestWrapper {
...
+      }
+    }
+    return config;
+  }
+
+  public enableRetries(retryConfig?: RetryOptions) {
+    this.axiosInstance.defaults.raxConfig = {
+      instance: this.axiosInstance,
+    };
+    this.retryConfig = RequestWrapper.getRetryConfig(this.axiosInstance, retryConfig);
```
padamstx, line 395: suggested name changes:
```suggestion
    this.raxConfig = RequestWrapper.getRaxConfig(this.axiosInstance, retryOptions);
```

padamstx (commented):
I left a bunch of picky comments about potentially changing variable names.   Sorry for that, but I think it will help us maintain our own sanity if we're super clear about the two different types of "retry config".

lib/request-wrapper.ts
barrett-schonefeld: Yes, I should not need the value after `??`.

lib/base-service.ts
barrett-schonefeld: This should be `RetryOptions`. I created an interface and need to update the comment here.

lib/request-wrapper.ts
barrett-schonefeld: I think that's a good idea!

lib/request-wrapper.ts
barrett-schonefeld: Done!

lib/request-wrapper.ts
barrett-schonefeld: Done!

lib/request-wrapper.ts
```diff
@@ -346,6 +368,40 @@ export class RequestWrapper {
...
+        config.retry = retryConfig.maxRetries;
+      }
+      if (typeof retryConfig.maxRetryInterval === 'number') {
+        config.maxRetryDelay = retryConfig.maxRetryInterval;
+      }
+    }
+    return config;
+  }
+
+  public enableRetries(retryConfig?: RetryOptions) {
```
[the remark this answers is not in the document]
barrett-schonefeld: Done.

lib/request-wrapper.ts
```diff
@@ -346,6 +368,40 @@ export class RequestWrapper {
...
+    this.axiosInstance.defaults.raxConfig = {
+      instance: this.axiosInstance,
+    };
+    this.retryConfig = RequestWrapper.getRetryConfig(this.axiosInstance, retryConfig);
+    this.retryInterceptorId = rax.attach(this.axiosInstance);
+  }
+
+  public disableRetries() {
+    if (this.retryInterceptorId) {
+      rax.detach(this.retryInterceptorId, this.axiosInstance);
```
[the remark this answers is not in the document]
barrett-schonefeld: Yes and retryInterceptorId

lib/request-wrapper.ts
```diff
@@ -36,10 +37,29 @@ import {
...
+ * Retry configuration options.
+ */
+export interface RetryOptions {
+  /**
+   * Maximum retries to attempt.
+   */
+  maxRetries?: number;
+
+  /**
+   * Ceiling for the retry delay (in ms) - delay will not exceed this value.
```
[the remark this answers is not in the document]
barrett-schonefeld: Fixed comment!

lib/request-wrapper.ts
```diff
@@ -346,6 +368,40 @@ export class RequestWrapper {
...
+      instance: axiosInstance,
+      backoffType: 'exponential',
+      checkRetryAfter: true,
+    };
+    if (retryConfig) {
+      if (typeof retryConfig.maxRetries === 'number') {
+        config.retry = retryConfig.maxRetries;
+      }
+      if (typeof retryConfig.maxRetryInterval === 'number') {
+        config.maxRetryDelay = retryConfig.maxRetryInterval;
```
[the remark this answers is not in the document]
barrett-schonefeld: Done!

lib/request-wrapper.ts
dpopp07: Not sure if this accomplishes what you want, but you could potentially use the "or" operator to make this conditional:
```
raxConfig: this.retryConfig || undefined,
```

All properties that are not set are technically `undefined` so if the `retryConfig` variable isn't set, then the result would be the same as the `raxConfig` not being set at all

dpopp07:
@barrett-schonefeld I sent this info on Slack but wanted to capture it here - the reason behind your `is not a function` problem is that the `RequestWrapper` class is mocked in the base service tests. See [this block of code](https://github.com/IBM/node-sdk-core/blob/ed85c17bada95dab0a9aaaf0560783a80a734b23/test/unit/base-service.test.js#L16-L19).

The code itself looks good, you will just need to update the mock for the tests accordingly. You can re-add your setter as well, you'll just need to do the same thing

test/unit/base-service.test.js
```diff
@@ -429,6 +429,22 @@ describe('Base Service', () => {
...
+    const config = {
+      maxRetries: 5,
+      maxRetryInterval: 100,
+    }
+    testService.enableRetries(config);
+
+    expect(testService.requestWrapperInstance.raxConfig.retry).toBe(5);
+    expect(testService.requestWrapperInstance.raxConfig.maxRetryDelay).toBe(100 * 1000);
+    expect(testService.requestWrapperInstance.raxConfig.backoffType).toBe('exponential');
+    expect(testService.requestWrapperInstance.raxConfig.checkRetryAfter).toBe(true);
```
dpopp07, line 446: You should test this actual behavior in the request wrapper tests. For this class, just add a test confirming that the request wrapper mock was called with the correct options. You can do the same for the disabling of retries.

lib/request-wrapper.ts
```diff
@@ -36,10 +37,29 @@ import {
...
 export class RequestWrapper {
-  private axiosInstance;
+  private axiosInstance: AxiosInstance;
+
+  private retryInterceptorId: number;
 
-  private compressRequestData: boolean;
+  private raxConfig: rax.RetryConfig;
+
+  public compressRequestData: boolean;
```
dpopp07, line 62: I liked your idea of keeping this private and adding a setter, now that we've figured out the "not a function" error

lib/request-wrapper.ts
```diff
@@ -240,6 +260,7 @@ export class RequestWrapper {
       headers,
       params: qs,
       data,
+      raxConfig: this.raxConfig,
```
dpopp07, line 263: So I don't actually think this needs to be conditional - if `raxConfig` is not set, this will just be undefined and it shouldn't do anything

dpopp07 (commented):
Looks great! I left a couple comments and I'll wait for tests to approve but the code changes and looking solid 👍

lib/request-wrapper.ts
barrett-schonefeld: Great, I'll use this approach and pass undefined when it's undefined

lib/request-wrapper.ts
barrett-schonefeld: Yes, changed it back

PR #158, opened 2021-08-31
perf: use maxContentLength -1 instead of Infinity

ricellis:
I stumbled across https://github.com/axios/axios/issues/2829 when investigating a performance issue in our SDK that extends the core. The performance issue described in that issue can be avoided by using `-1` instead of `Infinity` to set a preference for having no `maxContentLength`.

Since the purpose of `Infinity` here is to have no limit, using `-1` should make no difference to the expected behaviour, but it does avoid the performance problem (at least until such time as axios have a fix for it). In the profiling of my particular scenario having `maxContentLength` enabled was causing 25% of all ticks to be spent in `Buffer` `concat` which was having a dramatic performance impact.

##### Checklist

- [x] tests are included - updated default value test

### DIFF
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -70,7 +70,7 @@ export class RequestWrapper {
     // to 'application/x-www-form-urlencoded'. This causes problems, so overriding the
     // defaults here
     const axiosConfig: AxiosRequestConfig = {
-      maxContentLength: Infinity,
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM

dpopp07 (approved):
Great catch, thanks for the PR! Looks good 👍

PR #161, opened 2021-09-09
chore: update dependencies

barrett-schonefeld:
Screen Shot 2021-09-09 at 12 35 08 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,20 +19,20 @@
       "dev": true
     },
     "@babel/core": {
-      "version": "7.15.0",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.15.0.tgz",
-      "integrity": "sha512-tXtmTminrze5HEUPn/a0JtOzzfp0nk+UEXQ/tqIJo3WDGypl/2OFQEMll/zSFU8f/lfmfLXvTaORHF3cfXIQMw==",
+      "version": "7.15.5",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -2385,18 +2385,17 @@
       }
     },
     "conventional-commits-parser": {
-      "version": "3.2.1",
-      "resolved": "https://registry.npmjs.org/conventional-commits-parser/-/conventional-commits-parser-3.2.1.tgz",
-      "integrity": "sha512-OG9kQtmMZBJD/32NEw5IhN5+HnBqVjy03eC+I71I0oQRFA5rOgA4OtPOYG7mz1GkCfCNxn3gKIX8EiHJYuf1cA==",
+      "version": "3.2.2",
```
barrett-schonefeld, line 2388: Looks good

padamstx:
> Screen Shot 2021-09-09 at 12 35 08 PM

yes, but I've found that `npm audit` didn't pick up the original vulnerabilities so I'm not sure about its credibility :)

padamstx (approved):
Looks good, but you'll need to use a commit message that starts with something like `fix(build): ...` to trigger a new patch release of the node core library so that users can immediately get the fixed version.

Edit: actually after thinking about this a little more and reviewing your node template PR, I guess node SDKs would not actually need to pull in a new version of the node core to fix this issue since they have to go through the same steps, etc.

PR #165, opened 2021-09-22
fix: support POST for retry-axios

mojito317:
Retry does not emit when `POST` request is called in the Node SDK. This is happening because the configuration of retry-axios defaults to the `['GET', 'HEAD', 'OPTIONS', 'DELETE', 'PUT']` verbs ([source](https://github.com/JustinBeckwith/retry-axios#usage)).

I added a test to make sure retry is happening when `POST` is called. This test would fail without the augment.

##### Checklist

- [x] tests are included


dpopp07 (approved):
Looks good! 👍 

Thanks for the PR. I will wait for @barrett-schonefeld to look at this as well and then I will merge

PR #166, opened 2021-09-29
build: update vulnerable dependencies

barrett-schonefeld:
I updated to the latest version of `Jest`, which removes the dependency on `set-value`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6245,12 +6245,10 @@
       "dependencies": {
         "@gar/promisify": {
           "version": "1.1.2",
-          "bundled": true,
           "dev": true
         },
         "@npmcli/arborist": {
           "version": "2.8.3",
-          "bundled": true,
           "dev": true,
           "requires": {
             "@npmcli/installed-package-contents": "^1.0.7",
@@ -6288,12 +6286,10 @@
         },
         "@npmcli/ci-detect": {
           "version": "1.3.0",
-          "bundled": true,
           "dev": true
         },
         "@npmcli/config": {
           "version": "2.3.0",
-          "bundled": true,
           "dev": true,
           "requires": {
             "ini": "^2.0.0",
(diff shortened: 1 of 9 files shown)


padamstx:
@barrett-schonefeld The [two dependabot alerts](https://github.com/IBM/node-sdk-core/security/dependabot) that we're seeing in the node core indicate that the following dependency upgrades need to be made:
1. semver-regex  >=3.1.3  (there's actually a PR open for this change)
2. set-value >=4.0.1 (that would be a major version change, so not sure if that would break anything).

I don't think your proposed changes address either of those, do they?   In fact, I'm not really seeing any version # changes at all... looks like the only changes are the removal of the "bundled" lines from package-lock.json.

I'll go ahead and merge in the semver-regex-related PR that was opened up by dependabot automatically, as that will at least address one of the vulnerabilities.

barrett-schonefeld:
> I don't think your proposed changes address either of those, do they?

Not initially, but now I updated `Jest` to the latest, which removes `set-value` as a dependency.

> I'll go ahead and merge in the semver-regex-related PR that was opened up by dependabot automatically, as that will at least address one of the vulnerabilities.

I rebased after you merged the `semver-regex` work, so both vulnerabilities should be resolved.

barrett-schonefeld:
> Not initially, but now I updated `Jest` to the latest, which removes `set-value` as a dependency.

I need to update some of the tests to work with the latest version of `Jest`.

barrett-schonefeld:
Removed the use of done in async tests because `Jest@27` does not allow tests to both utilize `done` and return a promise.

For more information on this change, see this [Jest issue](https://github.com/facebook/jest/issues/10529).

padamstx (approved):
LGTM

dpopp07 (approved):
Looks good! Thanks for doing this, Barrett. Glad to finally be rid of the redundant `done()` calls 🙂

PR #177, opened 2021-11-11
expose as ES6 in addition to UMD

CarstenLeue:
Added a build step to build ES6 output in addition to UMD style output. The effect is that webpack based tools can now treeshake the component.

##### Checklist

- [ ] tests are included


CarstenLeue:
Example of the effect of the es6 compile for an application that only contains this snippet (https://github.com/CarstenLeue/node-sdk-core-tree-shaking):

```typescript
import { getAuthenticatorFromEnvironment } from 'ibm-cloud-sdk-core';

const auth = getAuthenticatorFromEnvironment('SECRETS_MANAGER_API');
console.log('auth', auth);
```

with tree shaking:
image

without tree shaking (i.e. before this PR):
image

Notice that the tree-shaken version does no longer contain the expect, jest and babel dependencies.

Both versions are still pretty large, but this is a consequence of using the `mime-types` and `tough-cookie` dependencies. If you are interested in providing a compact layer, you might want to revisit these dependencies.

CarstenLeue:
@dpopp07 when building the `ibm-cloud-sdk-core` via `npm run all` I notice that the `dist` folder is populated but there is no `package.json` inside that dist folder. Maybe this gets added via the final publish step. 
It's important that in the `package.json` that finally gets published you find the `module` and `sideEffects` entry (https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)

dpopp07:
@CarstenLeue I would support being more explicit in adding the `package.json` to the `dist` folder if you want. You could add a copy step to the build step, we've done this in other packages where we needed to be more explicit, e.g.

`"build": "tsc && cp package.json dist/",`

CarstenLeue:
@dpopp07 sure. I have added a step to copy the `package.json` into the dist folder. I am not using `cp` since that does not work cross platform. Also I am removing the sections from the copied `package.json` that should not be published anyway.

dpopp07 (commented):
This looks great! Just have one question to address before approving

dpopp07:
@CarstenLeue is there anything we need to do manually to support the new documentation related additions in this PR?

etc/ibm-cloud-sdk-core.api.md
```diff
@@ -0,0 +1,458 @@
+## API Report File for "ibm-cloud-sdk-core"
```
padamstx, line 1: Perhaps this one also should not be tracked in git???

es/tsdoc-metadata.json
```diff
@@ -0,0 +1,11 @@
+// This file is read by tools that parse documentation comments conforming to the TSDoc standard.
+// It should be published with your NPM package.  It should not be tracked by Git.
```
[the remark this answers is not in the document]
CarstenLeue: let me remove it

CarstenLeue:
> @CarstenLeue is there anything we need to do manually to support the new documentation related additions in this PR?

@dpopp07 maybe not manually but there should be a step in the automation that publishes the npm package. This step should also write the generated doc files to the `gh-pages` branch, e.g. like so https://docs.travis-ci.com/user/deployment/pages/

But I am lacking information how the actual build integration is working

etc/ibm-cloud-sdk-core.api.md
CarstenLeue: This file is supposed to be tracked https://api-extractor.com/pages/overview/demo_api_report/
It contains the description of the public API surface. Tracking it allows you to detect intended or unintended changes to the public API.

padamstx (approved):
LGTM

PR #180, opened 2021-12-05
fix: add new validateParams function

padamstx:
This commit adds the "validateParams" function
to the node core.  This function will be used
by the operation code generated by the Node
SDK generator to validate the parameters found
within a "params" object.  It will check for
required parameters that are not present, and also
check to make sure that all parameters that ARE present
are valid (i.e. the param name is among the expected
valid parameter names for the operation).

##### Checklist

- [x] tests are included


test/unit/validate-params.test.js
```diff
@@ -0,0 +1,115 @@
...
+    expect(validateParams({}, undefined, undefined)).toBeNull();
+  });
+
+  it('should return null if params is null and requiredParams and allParams are empty lists', () => {
+    expect(validateParams(null, [], [])).toBeNull();
+  });
+
+  it('should return null if params is an empty object and requiredParams and allParams are null', () => {
+    expect(validateParams({}, null, null)).toBeNull();
+  });
```
dpopp07, line 50: So thorough with the null/undefined/empty checks! 😄

dpopp07 (approved):
Looks great! 👍

PR #184, opened 2021-12-17
fix: avoid Buffer re-encode during compression

ricellis:
This is similar to the problem fixed by #182 but in this case for `Buffer` request body data instead of a `ReadableStream`.

The fix in this case is if the `data` is already a `Buffer` then just compress it, don't convert it to a `string` and back to a `Buffer` again before compressing it.
 
I added a test with `Buffer` input. I chose the input carefully to ensure that if it was assumed to be `UTF-8` encoded it would not be the same when decompressed and added an additional assertion for the decompressed data to match the original.

##### Checklist

- [x] tests are included


dpopp07 (approved):
Looks good! Thanks for the clever test here 👍

PR #186, opened 2022-01-04
build: use npm8 and package-lock v2

padamstx:
This PR migrates the package-lock.json file to v2 and modifies travis.yml to use npm v8 during the travis build.

##### Checklist

- [ ] tests are included


dpopp07 (approved):
I don't think you need to keep the package.json changes but I suppose it's harmless either way. Looks good! 👍

padamstx:
> I don't think you need to keep the package.json changes but I suppose it's harmless either way. Looks good! +1

Good catch...  I guess if we were to ever move the package to be scoped, then we already have that in place :)

Issue #179, opened 2021-11-24
Bug in DEBUG ?


huineng:
when i add my process.env.DEBUG variable to "ibm-cloud-sdk-core:info" i not only get the info items i also get an ERROR log , while this is not an error

"@ibm-cloud/cloudant": "^0.0.21",
"typescript": "^4.5.2",
"webpack": "^5.64.3",

node v16.13.0

```
  ibm-cloud-sdk-core:info Credential file does not exist. Will not be used +0ms
  ibm-cloud-sdk-core:info Credential file does not exist. Will not be used +2ms
  ibm-cloud-sdk-core:info Credential file does not exist. Will not be used +3ms
  ibm-cloud-sdk-core:error {
  ibm-cloud-sdk-core:error   status: 200,
  ibm-cloud-sdk-core:error   statusText: 'OK',
  ibm-cloud-sdk-core:error   headers: {
  ibm-cloud-sdk-core:error     'cache-control': 'must-revalidate',
  ibm-cloud-sdk-core:error     connection: 'close',
  ibm-cloud-sdk-core:error     'content-length': '208',
  ibm-cloud-sdk-core:error     'content-type': 'application/json',
  ibm-cloud-sdk-core:error     date: 'Wed, 24 Nov 2021 20:29:36 GMT',
  ibm-cloud-sdk-core:error     server: 'CouchDB/2.3.1 (Erlang OTP/19)',
  ibm-cloud-sdk-core:error     'x-couch-request-id': 'a94debd32c',
  ibm-cloud-sdk-core:error     'x-couchdb-body-time': '0'
  ibm-cloud-sdk-core:error   },
....
ibm-cloud-sdk-core:error   data: {
  ibm-cloud-sdk-core:error     couchdb: 'Welcome',
  ibm-cloud-sdk-core:error     version: '2.3.1',
  ibm-cloud-sdk-core:error     git_sha: 'c298091a4',
  ibm-cloud-sdk-core:error     uuid: '3fd2dc04a0c4b62f04ff84157d3bca8d',
  ibm-cloud-sdk-core:error     features: [ 'pluggable-storage-engines', 'scheduler' ],
  ibm-cloud-sdk-core:error     vendor: { name: 'The Apache Software Foundation' }
  ibm-cloud-sdk-core:error   }
  ibm-cloud-sdk-core:error } +0ms
[...]

PR #188, opened 2022-01-14
fix: avoid errors during logging of requests/responses

padamstx:
Fixes https://github.com/IBM/node-sdk-core/issues/179

This commit slightly modifies how requests and responses
are logged when debug logging is enabled to avoid
any JSON marshalling type errors.
We no longer use JSON.stringify() to format
the request and response objects, as they
actually aren't needed because the
logger.debug() and logger.error() functions
can handle objects just fine and do a decent job
of pretty-printing them.

##### Checklist

- [ ] tests are included

### DIFF
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -132,22 +132,12 @@ export class RequestWrapper {
       this.axiosInstance.interceptors.request.use(
         (config) => {
           logger.debug('Request:');
-          try {
-            logger.debug(JSON.stringify(config, null, 2));
-          } catch {
-            logger.error(config);
-          }
-
+          logger.debug(config);
           return config;
         },
         (error) => {
           logger.error('Error: ');
-          try {
-            logger.error(JSON.stringify(error, null, 2));
-          } catch {
-            logger.error(error);
-          }
-
+          logger.error(error);
           return Promise.reject(error);
         }
       );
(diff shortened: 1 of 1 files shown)


lib/request-wrapper.ts
```diff
@@ -132,45 +132,25 @@ export class RequestWrapper {
       this.axiosInstance.interceptors.request.use(
         (config) => {
           logger.debug('Request:');
-          try {
```
padamstx, line 135: I simply removed the try/catch blocks since logger.debug would not throw an exception.

lib/request-wrapper.ts
```diff
@@ -132,45 +132,25 @@ export class RequestWrapper {
...
         },
         (error) => {
           logger.error('Error: ');
-          try {
-            logger.error(JSON.stringify(error, null, 2));
-          } catch {
-            logger.error(error);
-          }
-
+          logger.error(error);
```
padamstx, line 140: I continued using logger.error() in this case instead of changing this to logger.debug(), because this is logging the request after some sort of error has occurred.
To be honest, I'm not quite sure what sort of error might cause trigger this error path while trying to log the request, because not even a connection-refused type error would trigger it (I tried :) ).

lib/request-wrapper.ts
```diff
@@ -132,45 +132,25 @@ export class RequestWrapper {
...
       this.axiosInstance.interceptors.response.use(
         (response) => {
           logger.debug('Response:');
-          try {
-            logger.debug(JSON.stringify(response, null, 2));
-          } catch {
-            logger.error(response);
-          }
-
+          logger.debug(response);
```
padamstx, line 148: This response logging uses the same pattern as above for the request logging.

lib/request-wrapper.ts
ricellis: I couldn't trigger this either. I suppose one possibility is in the case of multiple interceptors and one earlier in the chain rejects, but regardless IMO it doesn't impact the correctness of this fix.

ricellis (approved):
LGTM
I also tested locally and it looks fixed.

dpopp07 (approved):
Looks good! 👍

PR #191, opened 2022-02-10
fix: bump follow-redirects to avoid vuln

padamstx:
This PR bumps our follow-redirects version to avoid a vulnerability.


dpopp07 (approved):
Looks good! 👍

PR #192, opened 2022-03-07
chore: update minimum Node version in package.json

pyrooka:
This PR updates the minimum Node version for this package
from v10 to v12. Node.js 10 is no longer supported ([see here](https://endoflife.date/nodejs)).
This change also means the minimum TLS version has been set to TLS v1.2 by default.
`tls.DEFAULT_MIN_VERSION` - Default: 'TLSv1.2' (Added in: v11.4.0) [Docs](https://nodejs.org/docs/latest-v12.x/api/tls.html#tls_tls_default_min_version)

##### Checklist

- [ ] tests are included


padamstx:
@ricellis Sorry, wrong PR :)

PR #194, opened 2022-03-11
update axios

dpopp07:
Updates `axios` to the latest version to resolve a vulnerability and bring in the newest fixes/features.

This also updates our documentation and typing around cookie support, as it is currently inaccurate.


lib/request-wrapper.ts
```diff
@@ -65,24 +65,10 @@ export class RequestWrapper {
...
-        post: {
-          'Content-Type': 'application/json',
-        },
-        put: {
-          'Content-Type': 'application/json',
-        },
-        patch: {
-          'Content-Type': 'application/json',
-        },
-      },
```
dpopp07, line 85: Moved to another point in the code due to typing changes in `axios`

package.json
```diff
@@ -63,22 +64,21 @@
...
     "prettier": "^2.3.0",
+    "rimraf": "^3.0.2",
     "semantic-release": "18.0.1",
-    "@masterodin/publisher": "^0.10.0",
     "typedoc": "^0.17.0",
-    "typescript": "~3.8.3",
-    "npm-run-all": "^4.1.5",
-    "package-json-reducer": "^1.0.18",
-    "rimraf": "^3.0.2"
+    "typescript": "~3.8.3"
```
dpopp07, line 74: I guess NPM didn't like how we were ordering these!

lib/request-wrapper.ts
```diff
@@ -107,6 +93,13 @@ export class RequestWrapper {
 
     this.axiosInstance = axios.create(axiosConfig);
 
+    // axios sets the default Content-Type for `post`, `put`, and `patch` operations
+    // to 'application/x-www-form-urlencoded'. This causes problems, so overriding the
+    // defaults here
+    ['post', 'put', 'patch'].forEach((op) => {
```
padamstx, line 99: I assume these defaults are only used if the request does not contain the Content-Type header, but we'd expect Content-Type to be set on all requests invoked via an SDK (that have a request body).  Is that right?

padamstx (approved):
LGTM

lib/request-wrapper.ts
dpopp07: Yes, that's correct. Axios just wants to use `application/x-www-form-urlencoded` as the default and `application/json` is a better fallback assumption. I don't remember what the scenario was that involved the default being used and us discovering this problem but the solution has been in the code for a while

PR #205, opened 2022-07-25
fix: update file-type to resolve vulnerability

### LINKED ISSUE #204: CVE-2022-36313 found in file-type
https://security.snyk.io/vuln/SNYK-JS-FILETYPE-2958042

this project imports an old version of [file-type](https://github.com/IBM/node-sdk-core/blob/main/package.json#L86) and now a vulnerability has been discovered.

dpopp07:
Resolves #204 

The CVE states that v16.5.4 is clean, so I upgraded to that and modified the code accordingly. It led to some function changes from sync to async but all functions that should really only be used internally. That said, I think these functions are technically exported as part of the public API, so we should evaluate if this should considered a breaking change or a necessary fix of broken (vulnerable) code.

I wanted to upgrade to v17 but that caused a number of TypeScript errors from the `file-type` internal code. I may open an issue there to see if I can find out why it's happening.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "npm-run-all": "^4.1.5",
     "object.assign": "~4.1.0",
     "package-json-reducer": "^1.0.18",
-    "prettier": "^2.3.0",
+    "prettier": "~2.3.0",
     "rimraf": "^3.0.2",
     "semantic-release": "19.0.3",
     "typescript": "~3.8.3"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "npm-run-all": "^4.1.5",
     "object.assign": "~4.1.0",
     "package-json-reducer": "^1.0.18",
-    "prettier": "^2.3.0",
+    "prettier": "~2.3.0",
```
dpopp07, line 69: While working on this, I ran into an issue where a later version of prettier caused a whole mess of TypeScript issues. See [this issue](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/60314) for more info. We need to keep this version here until we can upgrade our TypeScript version (which has always caused other issues).

padamstx (approved):
LGTM

padamstx:
I restarted the bulids for this PR because the original builds were canceled by travis due to the build credits issue.

PR #206, opened 2022-07-28
fix: correct AuthenticatorInterface type

### LINKED ISSUE #176: Authenticator returns `Promise<void | Error>`
The authenticator interface has the following syntax:

```typescript
public authenticate(requestOptions: AuthenticateOptions): Promise<void | Error>
```
https://github.com/IBM/node-sdk-core/blob/main/auth/authenticators/authenticator.ts

Normally a Promise handles errors via its exception channel and regular values via its data channel. The type parameter identifies the data channel, promises do not have a way to specify the type of the exceptional channel.

Is it indeed intended to pass an error on the data channel? If so, why?
If not, the signature should be:

```typescript
public authenticate(requestOptions: AuthenticateOptions): Promise<void>
```

dpopp07:
The interface incorrectly defined the data channel as "void or Error" when it
should only be "void", as Errors are sent in a different channel. This updates
the type to be correct. It should not constitute a breaking update as we
already defined authenticators with the new type in this project.

Resolves #176

Signed-off-by: Dustin Popp <dpopp07@gmail.com>

### DIFF
--- a/auth/authenticators/authenticator-interface.ts
+++ b/auth/authenticators/authenticator-interface.ts
@@ -44,7 +44,7 @@ export interface AuthenticatorInterface {
    * @param {object.<string, string>} requestOptions.headers The headers the
    *   authentication information will be added to.
    */
-  authenticate(requestOptions: AuthenticateOptions): Promise<void | Error>;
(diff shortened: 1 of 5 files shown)


padamstx (approved):
LGTM

PR #207, opened 2022-07-28
feat: add `setDefaultHeaders` method on base service class

dpopp07:
This change adds a setter for the default headers stored on the base service.
These headers are sent with every SDK request. All of the other language cores
supported setting these values post-construction so this brings Node to parity.

Signed-off-by: Dustin Popp <dpopp07@gmail.com>


padamstx (approved):
LGTM

PR #211, opened 2022-10-21
fix: allow for synchronous building of file requests

dpopp07:
This issue was caused by an oversight in adding an `await` statement within a `forEach` loop, which will result in asynchronous operations happening without the synchronous control you'd expect from async/await.

This solves the issue by simply converting the loops to "for of" loops.

This caused a subtle server error so unfortunately, I can't think of a great test case for it. I should've known not to put an `await` call within a `forEach` though. Logically, this is a simple refactor, but practically it makes the code behave in the way it was actually intended to.

cc @apaparazzi0329

### DIFF
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -173,22 +173,19 @@ export class RequestWrapper {
 
     // Form params
     if (formData) {
-      Object.keys(formData).forEach((key) => {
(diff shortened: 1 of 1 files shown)


dpopp07:
Seeing some linting failures (related to async/await within loops, ironically). I'll get those sorted

padamstx (approved):
Looks good except for the build failures :)

Issue #209, opened 2022-10-05
 upgrade axios-cookiejar-support to support axios 1.0


huineng:
in your package json you are using  axios-cookiejar-support ^1.0.0

this version is already at 4.0.3 https://github.com/IBM/node-sdk-core/blob/main/package.json

today axios itself upgraded to version 1 https://github.com/axios/axios/releases/tag/v1.0.0

and when using axios v1, node_sdk is no longer working

example of an error 
```
Cannot find module 'axios/lib/defaults' from 'node_modules/axios-cookiejar-support/lib/index.js'

Require stack:
      node_modules/axios-cookiejar-support/lib/index.js
      node_modules/ibm-cloud-sdk-core/lib/request-wrapper.js
      node_modules/ibm-cloud-sdk-core/auth/token-managers/token-manager.js
      node_modules/ibm-cloud-sdk-core/auth/token-managers/jwt-token-manager.js
      node_modules/ibm-cloud-sdk-core/auth/token-managers/iam-request-based-token-manager.js
      node_modules/ibm-cloud-sdk-core/auth/token-managers/iam-token-manager.js
      node_modules/ibm-cloud-sdk-core/auth/token-managers/index.js
      node_modules/ibm-cloud-sdk-core/auth/authenticators/cloud-pak-for-data-authenticator.js
      node_modules/ibm-cloud-sdk-core/auth/authenticators/index.js
      node_modules/ibm-cloud-sdk-core/auth/index.js
      node_modules/ibm-cloud-sdk-core/lib/base-service.js
      node_modules/ibm-cloud-sdk-core/index.js
      node_modules/@ibm-cloud/cloudant/cloudant/v1.js
      node_modules/@ibm-cloud/cloudant/index.js
```

ps i might need to check with that library if it's also compatible with axios 1.0

thanks

is it possible to make the library compatible with axios 1.0 and the latest version of axios-cookiejar-support

thanks

andyedwardsibm:
Hitting this as well.  As a simple recreate...

```
$# npm install axios@1.0.0 ibm-cloud-sdk-core

$# npm ls axios
recreate@ /tmp/recreate
├── axios@1.0.0
└─┬ ibm-cloud-sdk-core@3.1.0
  ├─┬ axios-cookiejar-support@1.0.1
  │ └── axios@1.0.0 deduped
  ├── axios@0.26.1
  └─┬ retry-axios@2.6.0
    └── axios@1.0.0 deduped

[...]

PR #213, opened 2022-11-14
[WIP] attempt upgrade to axios 1.x

ricellis:
An attempt to upgrade to axios 1.x.

Commit summary:
* 8d81238638a466422446d02267e8f0512c56c2db - baseline, can be dropped, just happened to be what I had locally was a different version.
* 890e8b659bfbabe3186f2e8aa6897d62cfbb9a68 - update the `package.json` depenencies
    * build(deps): update axios to 1.2.0
    * This also requires new versions of:
        * `typescript` - to support the axios 1.x types
        * `jest` - to support the axios 1.x module
        * `axios-cookiejar-support` for axios 1.x module compatibility
    * Note: 1.2.0-alpha.1 has fixes to axios request headers that stop the retry tests from failing. Unclear on when it may move from alpha to a real release.
* b18877952cd4e097d3b20f4d587297f5ed843a4f - fix axios and cookiejar support usage and tests
    * fix: axios 1.x in request-wrapper
    * Use `paramsSerializer` legacy compatibility `serialize` function with existing stringify function.
    * Correct axios-cookiejar-support import and intialization.
    * Provide backwards compatibility for `jar: true` option.
    * Correct mocking in request-wrapper.test.js.
    * Remove `withCredentials` from cookiejar.test.js.
    * Set `synchronous: true` on mock cookie jar.
    * Add new test for usage of real tough-cookie CookieJar.
* 2e4618b386c1bc73a86c039f4b2184277452572c - correct some other test errors I saw locally
    * build-request-file-object.test.js and get-content-type.test.js both make use of streams with non-existent paths to check errors. Possibly since upgrading jest these cause suites to fail with e.g.
    ```
    Error: ENOENT: no such file or directory, open '/fake/path/custom-name.env'
    Emitted 'error' event on ReadStream instance at:
        at emitErrorNT (node:internal/streams/destroy:157:8)
        at emitErrorCloseNT (node:internal/streams/destroy:122:3)
        at processTicksAndRejections (node:internal/process/task_queues:83:21) {
      errno: -2,
      code: 'ENOENT',
      syscall: 'open',
      path: '/fake/path/custom-name.env'
    ```
    * Since the errors are expected for these tests add a no-op handler to handle the error events without a warning

##### Checklist

- [x] tests are included

### DIFF
--- a/test/unit/build-request-file-object.test.js
+++ b/test/unit/build-request-file-object.test.js
@@ -61,7 +61,8 @@ describe('buildRequestFileObject', () => {
     });
 
     it('should handle file object with a readable stream as its `value`', async () => {
-      const fileStream = fs.createReadStream(filepath);
+      // Note add an on error handler to avoid unhandled error events
+      const fileStream = fs.createReadStream(filepath).on('error', () => {});
       fileStream.path = `/fake/path/${customName}`;
 
       const fileParams = {
@@ -76,7 +77,8 @@ describe('buildRequestFileObject', () => {
     });
 
     it('should handle path property being a buffer', async () => {
-      const fileStream = fs.createReadStream(filepath);
+      // Note add an on error handler to avoid unhandled error events
+      const fileStream = fs.createReadStream(filepath).on('error', () => {});
       fileStream.path = Buffer.from(`/fake/path/${customName}`);
 
       const fileParams = {
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 
 node_js:
-- 12
 - 14
+- 16
 
 cache:
   npm: false
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -19,7 +19,7 @@
 import axios, { AxiosInstance, AxiosRequestConfig } from 'axios';
 import * as rax from 'retry-axios';
 
-import axiosCookieJarSupport from 'axios-cookiejar-support';
+import { wrapper as axiosCookieJarSupport } from 'axios-cookiejar-support';
 import extend from 'extend';
 import FormData from 'form-data';
 import { OutgoingHttpHeaders } from 'http';
@@ -27,6 +27,7 @@ import { Agent } from 'https';
 import isStream from 'isstream';
 import { stringify } from 'querystring';
 import { gzipSync } from 'zlib';
+import { CookieJar } from 'tough-cookie';
 import {
   buildRequestFileObject,
   isEmptyObject,
@@ -104,8 +105,14 @@ export class RequestWrapper {
     if (axiosOptions.jar) {
       axiosCookieJarSupport(this.axiosInstance);
 
-      this.axiosInstance.defaults.withCredentials = true;
-      this.axiosInstance.defaults.jar = axiosOptions.jar;
+      // axios-cookiejar-support < 2.x accepted a boolean
+      // some SDKs might be using that, so create a default
+      // CookieJar if detecting true.
+      if (axiosOptions.jar === true) {
+        this.axiosInstance.defaults.jar = new CookieJar();
+      } else {
+        this.axiosInstance.defaults.jar = axiosOptions.jar;
+      }
     }
 
     // get retry config properties and conditionally enable retries
@@ -249,7 +256,7 @@ export class RequestWrapper {
       data,
       raxConfig: this.raxConfig,
       responseType: options.responseType || 'json',
-      paramsSerializer: (params) => stringify(params),
+      paramsSerializer: { serialize: (params) => stringify(params) },
     };
 
     return this.axiosInstance(requestParams).then(
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,8 +13,8 @@
         "@types/isstream": "^0.1.0",
         "@types/node": "~10.14.19",
         "@types/tough-cookie": "^4.0.0",
-        "axios": "^0.26.1",
-        "axios-cookiejar-support": "^1.0.0",
+        "axios": "1.2.0",
+        "axios-cookiejar-support": "4.0.3",
         "camelcase": "^5.3.1",
         "debug": "^4.1.1",
         "dotenv": "^6.2.0",
@@ -56,7 +56,7 @@
         "eslint-plugin-jsdoc": "^34.6.3",
         "eslint-plugin-node": "^9.0.0",
         "eslint-plugin-prettier": "^3.0.1",
-        "jest": "^27.2.4",
+        "jest": "^29.3.1",
         "nock": "^13.1.2",
         "npm-run-all": "^4.1.5",
         "object.assign": "~4.1.0",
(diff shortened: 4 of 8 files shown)


ricellis:
AFAICT the travis failures here are caused by `npm ci` failing to resolve the dep tree because the `1.2.0-alpha.1` appears not to semantically satisfy cookiejar support's `>=0.20.0`.
I could switch to `1.1.3` and demonstrate all but the retry test passing until axios `1.2.0` solidifies if that is preferred.

ricellis:
>However, it's a good confirmation on everything I'd worked on and you've also solved some things that were still on my to-do list

Happy to hear that it's on the same lines, do cherry-pick or copy/paste as needed into your work. We can close this PR down np.

>I'll go through your PR today and think about the dependency issue with the retry tests and get back to you.

With any luck a non-alpha 1.2.0 will happen and make this problem go away. I think anything using retries will break without the type fixes in that version (not just the tests).

>I also need to do some tests for compatibility with generated code and see if any of this constitutes breaking changes for the core

Yeah I wasn't sure whether we could get away with dropping the boolean on the `jar` option, but it seemed not a big deal to just make one to retain compatibility.

I'm not entirely sure on side-effects of the typescript version upgrade - I was definitely just focussed on getting axios working.

I do think this might not work well with Node 12, but that is out of support now anyway. If the engine does move then I'd also advocate for updating the version of the `@types/node` from `10` to `14` to match although I don't think it is strictly necessary usage wise atm.

dpopp07:
> Happy to hear that it's on the same lines, do cherry-pick or copy/paste as needed into your work. We can close this PR down np.

I think I'd rather just keep your PR - it's in better shape than mine!

> With any luck a non-alpha 1.2.0 will happen and make this problem go away. I think anything using retries will break without the type fixes in that version (not just the tests).

I think you're right. I ran into this problem as well. We can monitor for a release but the timeline might not be as soon as we'd like.

> Yeah I wasn't sure whether we could get away with dropping the boolean on the jar option, but it seemed not a big deal to just make one to retain compatibility.

I think I'd just dropped it in my local changes but I prefer what you've done here 👍 

> I'm not entirely sure on side-effects of the typescript version upgrade

This is what I'm most concerned about but I had done some tests on this subject already and things seemed like they should be fine. I just want to double check.

> I do think this might not work well with Node 12, but that is out of support now anyway

Yep, we have an issue on our internal board to move up the engines and update our Travis builds. We need to do that soon. Agreed that we should bump the types when we do that.

test/unit/build-request-file-object.test.js
```diff
@@ -61,7 +61,8 @@ describe('buildRequestFileObject', () => {
     });
 
     it('should handle file object with a readable stream as its `value`', async () => {
-      const fileStream = fs.createReadStream(filepath);
+      // Note add an on error handler to avoid unhandled error events
+      const fileStream = fs.createReadStream(filepath).on('error', () => {});
```
dpopp07, line 65: Good call. Should we fail the test on error?

ricellis: I did wonder whether this implied that the helper functions should attach an error handler in the stream cases so that they could reject for any stream errors, but the behaviour seemed to be relying on resolving as `null`. As it currently stands the function caller (sdk/user?) would have to attach an error listener to the stream they create or risk this unhandled error event (or missing the error).

I wasn't really sure whether it made sense to suppress the error in the helper code instead of the test code (i.e. leaving the current behaviour of resolving as null, but avoiding any chance of an unhandled errors) rather than pushing the responsibility down. Ofc it is a little scope-creep as it isn't really related to the axios upgrade at all, just something that happened to be detected now that wasn't before.

ricellis:
Just to add another note here on axios 1.2 and the retry issues.
The error I saw in the retry tests with 1.1.3 was [this invalid char in header content issue](https://github.com/axios/axios/issues/5089). When I upgraded to 1.2.0-alpha.1 the problem resolved. Noting that this [does not appear to be resolved by 1.2 for everyone](https://github.com/axios/axios/issues/5249#issuecomment-1318509074) so we may need to evaluate carefully.

ricellis:
[Axios 1.2.0 is out now](https://github.com/axios/axios/releases/tag/v1.2.0) so I've rebased using that instead of the `1.2.0-alpha.1` and as expected this is now green in Node 14, but incompatible with Node 12.

I pushed an additional commit to move to 14 minimum and a test matrix of 14 & 16 and it looks ok now.

dpopp07:
@ricellis Awesome - this looks great. Thanks again for all of your work on this. I will do a little testing to see how this plays with SDKs and ensure the changes are not "breaking" but I'd like to merge this soon

ricellis:
@dpopp07 is there a way to publish a snapshot build we could test too?

dpopp07:
@ricellis Yes, that's a great idea. I'll do that today

dpopp07:
@ricellis This branch can now be installed with the "beta" tag (`npm i ibm-cloud-sdk-core@beta`)

dpopp07:
@ricellis based on my testing, everything works fine but SDK developers will _need_ to update the `typescript` and `jest` versions for their respective projects. I'm guessing the same will go for developers/users of the actual SDKs. I'm not yet sure if I want to treat that as a breaking change or not

ricellis:
>seeing some problems in other downstream projects (specifically around axios-cookiejar-support and custom https agents). I'm looking into those.

AFAICT there is an incompatibility between newer versions of `axios-cookiejar-support` and `retry-axios` (and potentially some other interceptors/libs). I can't quite nail it down, but there is an [open issue](https://github.com/3846masa/axios-cookiejar-support/issues/602) and I've added my [thoughts](https://github.com/3846masa/axios-cookiejar-support/issues/602#issuecomment-1335074720) there.

In the meantime I'm going to try and rollback the `axios-cookiejar-support` version (I bumped it based on #209, but I'm not 100% sure it was necessary.

ricellis:
@dpopp07 I pushed another [branch](https://github.com/IBM/node-sdk-core/compare/ricellis/wip2-axios-1.x) with a commit 220c48eebbc3ab90c85b302858527bfd7d15d014 that removes the dependency on `axios-cookiejar-support` and uses an interceptor to interact with `tough-cookie` instead.

Given the availability of interceptors in axios I was looking for options that avoid the dependency that is giving us trouble although it is probably a bit rough around the edges, especially in terms of validating it works correctly when setting/getting multiple cookies for a single URL. I'm not sure how many other SDKs rely on cookie support, but if you want to consider making another beta from that branch/commit then I'd be happy to run our test pipelines and see if it works better for us.

dpopp07:
>  I'm not sure how many other SDKs rely on cookie support

I don't think that many do. I'll push a second beta with your changes shortly

ricellis:
@dpopp07 thanks for that.

The bad news is that I see a bunch of different failures using the `beta-2`; the good news is that they _weren't_ caused by my cookie change (I went back to `beta-1` and they were still present). Seems like the `beta-1` didn't get used in my SDK pipeline first time around (probably a bad merge on the package-lock) but was correctly used in my downstream lib pipelines which I guess explains why I saw more failures there. 

Anyway the "new" failures look like some kind of unicode or encoding problem - I'll invest some time into those and see if I can move this any closer to stability, but for now it is sadly too broken to proceed.

dpopp07:
@ricellis I updated the beta with your new changes

ricellis:
Thanks @dpopp07 - I couldn't see which commit made the `beta-3`, but it is still pulling axios `1.2.0` instead of `1.2.1`.

Please could you make one from [0f65ae1](https://github.com/IBM/node-sdk-core/commit/0f65ae19feb327c9aae27e1052f8cc19d767ea48)? I think that has the best chance of success. If we're still having issues after that I think we'll have to think of some kind of alternative or potentially wait for axios 1.x to stabilize further.

ricellis:
>Yep, I see what happened. My mistake. I just published the new beta from that commit

No worries, I appreciate the time you've been putting into releasing all these betas to help me test downstream and I didn't make it any easier with my own mistakes and having two separate branches!

`beta-4` definitely uses axios 1.2.1 and in general things look much better. The bad news is that I made an error in the cookie-support code. The good news is that when I use sed to patch that error in our pipelines there are no more test failures. I've got some soak tests running now.

PR #216, opened 2022-12-09
refactor: cookie support via interceptors

ricellis:
Attempts to upgrade to axios 1.x (#209/#213) have been impeded by incompatibilities with `axios-cookiejar-support`:
* 1.x version is [incompatible with newer axios versions](https://github.com/3846masa/axios-cookiejar-support/issues/310)
* 2.x+ versions are [incompatible with retries](https://github.com/3846masa/axios-cookiejar-support/issues/602)
Remove axios-cookiejar-support depdenency.

This PR:
* Removes `axios-cookiejar-support` dependency.
* Adds `cookie-support.ts` for providing axios interceptors to interact with `tough-cookie`.
* Updates the cookiejar tests for the above.

##### Checklist

- [x] tests are included
- [x] documentation is changed or added - N/A current docs about cookies remain valid

### DIFF
--- a/lib/cookie-support.ts
+++ b/lib/cookie-support.ts
@@ -16,7 +16,7 @@
 
 import extend from 'extend';
 import { AxiosRequestConfig, AxiosResponse } from 'axios';
-import { CookieJar } from 'tough-cookie';
+import { Cookie, CookieJar } from 'tough-cookie';
 import logger from './logger';
 
 export class CookieInterceptor {
@@ -61,9 +61,12 @@ export class CookieInterceptor {
       const cookies: string[] = response.headers['set-cookie'];
       if (cookies) {
         logger.debug(`CookieInterceptor: setting cookies in jar for URL ${response.config.url}.`);
-        cookies.forEach(async (cookie: string) => {
-          await this.cookieJar.setCookie(cookie, response.config.url);
-        });
+        // Write cookies sequentially by chaining the promises in a reduce
+        await cookies.reduce(
+          (cookiePromise: Promise<Cookie>, cookie: string) =>
+            cookiePromise.then(() => this.cookieJar.setCookie(cookie, response.config.url)),
+          Promise.resolve(null)
+        );
       } else {
         logger.debug('CookieInterceptor: no set-cookie headers.');
       }
--- a/lib/cookie-support.ts
+++ b/lib/cookie-support.ts
@@ -1,0 +1,75 @@
+/**
+ * (C) Copyright IBM Corp. 2022.
+ *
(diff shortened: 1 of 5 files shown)


lib/cookie-support.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  public async responseInterceptor(response: AxiosResponse) {
+    logger.debug('CookieInterceptor: intercepting response.');
+    if (response && response.headers) {
+      logger.debug('CookieInterceptor: checking for set-cookie headers.');
+      const cookies: string[] = response.headers['set-cookie'];
+      if (cookies) {
+        logger.debug(`CookieInterceptor: setting cookies in jar for URL ${response.config.url}.`);
+        cookies.forEach(async (cookie: string) => {
+          await this.cookieJar.setCookie(cookie, response.config.url);
```
dpopp07, line 65: There can be [some issues](https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop?page=1&tab=scoredesc#tab-top) with calling `async` functions within a `forEach` method. Are you intending to call `setCookie` sequentially?

dpopp07 (commented):
Looks great overall! Thanks for the contribution. I just have one comment about some `async` code. Might not need to change but wanted to check your intentions

lib/cookie-support.ts
ricellis: >Are you intending to call `setCookie` sequentially?

My real intention was to ensure that all the cookie writes had completed prior to the interceptor resolving and returning a response so that all the cookies would be readable by subsequent requests that happen after a cookie response.
I honestly hadn't given a lot of thought to whether the requests were running in parallel or sequentially and I don't know too much about the behaviour of the async cookie stores and whether they safely handle parallel writes. So given that in the majority of cases there is likely only a single cookie anyway and cookies are small writing them sequentially seems probably the safest thing to do. I've made an attempt at that in ca9c3045.

ricellis: >Are you intending to call setCookie sequentially?

Good spot, thanks. My main intention was to ensure that the cookies were written before the interceptor returned the response, such that any requests happening after that response would be able to read the cookies.

I honestly hadn't given a lot of thought to whether writes were happening sequentially or in parallel and I don't know a lot about the safety of parallel writes to asynchronous cookie stores (of which I guess there are potentially many implementations). With that safety in mind and given that it is likely that the cookies are small and there will be only a single cookie in the majority of cases I suspect that there won't be a huge performance penalty from writing sequentially. So I've made an attempt at changing to that in 3bf097b.

dpopp07 (approved):
Looks good! 👍

PR #217, opened 2022-12-09
Update to axios 1.2.1

ricellis:
Axios 0.x is not maintained any more and the 1.x stream contains some useful fixes (see also #209).
This PR is a cleaned up version of #213, but note that this PR requires #216 as a prerequisite ~(and since that is currently unmerged this PR also currently contains the commit from that PR)~. If there are `axios-cookiejar-support` changes then it may be possible to drop #216 and refactor this instead.

There are _potentially breaking changes_ here because:
* axios 1.x has a minimum of Node.js 14, so the minimum Node.js version here needs to increase (12.x is out of support)
* axios 1.x defines types with TS 4.x syntax so the minimum typescript version must update (this could impact downstream SDKs that are not yet using TS 4.x)

Commit summary:
* ~994a25c8441aa65b5c4f8eb6750f17b594e807c5 - from #216 _PREREQ_~
* [e0b86d2](https://github.com/IBM/node-sdk-core/pull/217/commits/e0b86d2c698c25dba6345cbdaa91d45684fed6c1) - _BREAKING_ update to Node.js 14 minimum as required by axios 1.x
* [8f16f08](https://github.com/IBM/node-sdk-core/pull/217/commits/8f16f0833ec65fd63fbb16127d623c779b90747d) - update axios dependency to `1.2.1`
* [fd3ba09](https://github.com/IBM/node-sdk-core/pull/217/commits/fd3ba096542c37b6cc32b428d5feb3d90d6dead3) - correct request-wrapper for axios 1.x
    * Use `paramsSerializer` legacy compatibility `serialize` function with existing stringify function.
    * Remove `withCredentials` from axios defaults.
    * Correct mocking in request-wrapper.test.js.
* [41729f2](https://github.com/IBM/node-sdk-core/pull/217/commits/41729f2f15ba5ccc59d6774ef65a6326d73f9eba) - _BREAKING_ update typscript to 4.x (needed for axios 1.x)
* [098dd62](https://github.com/IBM/node-sdk-core/pull/217/commits/098dd626e2be066b4bee6b6dcc1bdbf9d4dc9dbc) - update Jest to 29.x (needed for axios 1.x)
* [e884861](https://github.com/IBM/node-sdk-core/pull/217/commits/e8848618cc368686f73ca7be65d7f27dc847cd87) - correct other test errors
    * `build-request-file-object.test.js` and `get-content-type.test.js` both make use of streams with non-existent paths to check errors. Possibly since upgrading jest these cause suites to fail with e.g.
    ```
    Error: ENOENT: no such file or directory, open '/fake/path/custom-name.env'
    Emitted 'error' event on ReadStream instance at:
        at emitErrorNT (node:internal/streams/destroy:157:8)
        at emitErrorCloseNT (node:internal/streams/destroy:122:3)
        at processTicksAndRejections (node:internal/process/task_queues:83:21) {
      errno: -2,
      code: 'ENOENT',
      syscall: 'open',
      path: '/fake/path/custom-name.env'
    ```
    * Since the errors are expected for these tests add a no-op handler to handle the error events without a warning

##### Checklist

- [x] tests are included
- [x] documentation is changed or added - updated README for Node.js 14 minimum


ricellis:
Thanks @dpopp07 - I've updated this ready for another round of beta testing that definitely seems like a good plan!

dpopp07:
> I've updated this ready for another round of beta testing that definitely seems like a good plan!

Great - the new beta has been published! Let me know how testing goes. If all is well, I'll merge tomorrow

ricellis:
@dpopp07 our tests are still looking green with `beta-5` though given the nature of the change it would be more than a little reassuring to see feedback from any other SDKs too 😅

dpopp07:
> our tests are still looking green
Great

> given the nature of the change it would be more than a little reassuring to see feedback from any other SDKs too

Great point. We spent some time yesterday doing additional testing with the Platform Service SDKs and everything looks green on that end too. I think we're looking good to ship

dpopp07 (approved):
Looks great! Thanks for all of your work on this and for crafting your commit messages so well 👍

PR #218, opened 2022-12-09
feat: bump min node version to 14

padamstx:
This commit modifies min version of node supported by the project from 12 to 14, and modifies the travis build to use node versions 14 and 16.

Initially creating this as a draft PR for review purposes but we can hold off on merging this until after the axios upgrade work is finished (or perhaps include it in the same PR?).


padamstx:
Closing this PR since similar changes are already contained in https://github.com/IBM/node-sdk-core/pull/217

PR #221, opened 2022-12-20
docs: update changelog to include link to migration guide

dpopp07:
Give changelog a link to migration guide. Also, update dev dependency so it plays better with the new TS version.


package.json
```diff
@@ -49,7 +49,7 @@
     "@types/extend": "^3.0.1",
     "@types/lodash.isempty": "^4.4.6",
     "@types/mime-types": "^2.1.1",
-    "@typescript-eslint/eslint-plugin": "^4.24.0",
+    "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.24.0",
```
padamstx, line 53: Just wondering if we should keep the parser dependency at the same version as the eslint-plugin dependency?

padamstx (approved):
LGTM
Just had one question to consider.

package.json
dpopp07: Good call - we probably should. I'll push that change as well

PR #225, opened 2022-12-27
fix(auth): migrate to secure usage of jwt for token authentication

dpopp07:
There is a vulnerability in v8 of the `jsonwebtoken` dependency. This commit upgrades to v9 to resolve the vulnerability. Additionally, they made an effort in this version to discourage the less secure "decode" method in favor of the more secure "verify" method. This commit also refactors the code and tests to use the "verify" method.

### DIFF
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { decode } from 'jsonwebtoken';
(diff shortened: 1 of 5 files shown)


TannerS:
Thanks for this, please let us know an update, i think a lot of us have a short deadline to get this fixed, and this is a dep of another dep

apaparazzi0329 (approved):
There were no issues authenticating in the node sdk and the code looks good so you got my approval

padamstx (approved):
LGTM

PR #227, opened 2022-12-30
fix(auth): revert to using decode instead verify for jwt

dpopp07:
The move to `verify` was a bit shortsighted. It requires a public key or secret as an argument to verify the signed token. Typical usage of IBM's IAM service doesn't lend itself to this flow and the tokens returned don't have valid JWT signatures. This led to a runtime error everytime the core made an IAM request - a bit of a showstopping bug.

Additionally, we never had a goal of performing client-side validation of these tokens, we only decode them to determine the expiration time for usage in our refresh logic. The `decode` method is perfectly sufficient for that and indeed is called within the `verify` method anyways. Perhaps this flow will change in the future but this is all we need for now.

This reverts the logic to what we were using before but still using the safe version of the dependency. The new version makes the `decode` function read-only so I had to adjust our approach to mocking in the unit tests.

### DIFF
--- a/auth/token-managers/jwt-token-manager.ts
+++ b/auth/token-managers/jwt-token-manager.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { verify } from 'jsonwebtoken';
+import { decode } from 'jsonwebtoken';
 import logger from '../../lib/logger';
 import { TokenManager, TokenManagerOptions } from './token-manager';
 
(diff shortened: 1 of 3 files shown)


apaparazzi0329 (approved):
Sorry this is also on me. I pulled in the wrong changes when testing this with the node-sdk and I can confirm that I get the aforementioned error. This reversion should go in asap

PR #238, opened 2023-02-21
fix: return json error response body as object

dpopp07:
Error response bodies are currently returned as strings, even when the content is JSON. This is a bug - the SDK should return the result as an object. This commit adds logic to return the result as an object, converting it if necessary.

It also updates the error object to use the "result" field for the response body, in order to be consistent with the success path. The "body" field is still set for backward-compatibility but is marked as deprecated.

Additionally, for both success and error responses, if the content is JSON but the result cannot be parsed as an object, an exception will be raised, alerting the user that the service returned a malformed body.

cc @ricellis

### DIFF
--- a/lib/helper.ts
+++ b/lib/helper.ts
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2014, 2022.
+ * (C) Copyright IBM Corp. 2014, 2023.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -339,4 +339,15 @@ export function constructServiceUrl(
   });
 
   return formattedUrl;
+}
+
+/**
+ * Returns true if and only if "mimeType" is a "JSON-like" mime type
+ * (e.g. "application/json; charset=utf-8").
+ * @param mimeType - the mimeType string
+ * @returns true if "mimeType" represents a JSON media type and false otherwise
+ */
+export function isJsonMimeType(mimeType: string) {
+  logger.debug(`Determining if the mime type '${mimeType}' specifies JSON content.`);
+  return !!mimeType && /^application\/json(\s*;.*)?$/i.test(mimeType);
 }
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -30,6 +30,7 @@ import {
   isEmptyObject,
   isFileData,
   isFileWithMetadata,
+  isJsonMimeType,
   stripTrailingSlash,
 } from './helper';
 import logger from './logger';
@@ -266,7 +267,7 @@ export class RequestWrapper {
 
         // the other sdks use the interface `result` for the body
         // eslint-disable-next-line @typescript-eslint/dot-notation
-        res['result'] = res.data;
+        res['result'] = ensureJSONResponseBodyIsObject(res);
         delete res.data;
 
         // return another promise that resolves with 'res' to be handled in generated code
@@ -306,20 +307,19 @@ export class RequestWrapper {
 
       error.message = parseServiceErrorMessage(axiosError.data) || axiosError.statusText;
 
-      // some services bury the useful error message within 'data'
-      // adding it to the error under the key 'body' as a string or object
+      // attach the error response body to the error
       let errorBody;
       try {
-        // try/catch to handle objects with circular references
+        // try/catch to detect objects with circular references
         errorBody = JSON.stringify(axiosError.data);
       } catch (e) {
-        // ignore the error, use the object, and tack on a warning
+        logger.warn('Error field `result` contains circular reference(s)');
+        logger.debug(`Failed to stringify error response body: ${e}`);
         errorBody = axiosError.data;
-        errorBody.warning = 'Body contains circular reference';
-        logger.error(`Failed to stringify axiosError: ${e}`);
       }
 
-      error.body = errorBody;
+      error.result = ensureJSONResponseBodyIsObject(axiosError);
+      error.body = errorBody; // ** deprecated **
 
       // attach headers to error object
       error.headers = axiosError.headers;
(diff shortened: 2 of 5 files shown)


lib/request-wrapper.ts
```diff
@@ -545,3 +542,47 @@ function parseServiceErrorMessage(response: any): string | undefined {
...
+
+/**
+ * Check incoming headers to determine if the response body is meant to be JSON.
+ *
+ * @param headers - incoming headers attached to the response
+ * @returns true if content type indicates JSON, false otherwise
+ */
+function contentTypeIsJSON(headers: IncomingHttpHeaders): boolean {
+  const contentType = headers['content-type'] || '';
+  return contentType.toLowerCase() === 'application/json';
```
padamstx, line 554: I think whenever we check Content-Type for specific values, rather than doing a strict "===", we should tolerate mimetypes like `application/json; charset=utf8`.  In most situations (generator, cores, etc.) where we need to check Content-Type, we use a regex type comparison to account for the optional stuff at the end.

I did a quick check of the node core code, and I don't think we have anything like this place yet. Perhaps we could just introduce a utility function `isJsonMimetype()`.

test/unit/request-wrapper.test.js
```diff
@@ -801,22 +878,62 @@ describe('formatError', () => {
...
+    delete axiosErrorCopy.request;
+    const error = requestWrapperInstance.formatError(axiosErrorCopy);
     expect(error instanceof Error).toBe(true);
     expect(error.message).toBe('error in building the request');
   });
+
+  it('should convert string response bodies to objects if content is JSON', () => {
+    const newAxiosError = makeCopy(basicAxiosError);
+    newAxiosError.response.data = '{ "errorMessage": "some error" }';
+    expect(newAxiosError.response.headers['content-type']).toBe('application/json');
```
padamstx, line 901: let's try a content-type that has extra stuff (e.e. `application/json; charset=utf8`.

padamstx (approved):
Everything looks great, but I think we need to be tolerant of content-type values that might contain additional context stuff at the end.
We do this in a few other places (generator, cores, etc.) and we might want to also introduce this approach in the node core.  We probably have several other situations (either in the node core or other repos) where we should be doing regex matches instead of strict "===", so it's probably not critical that we do this now in this PR.   Just mentioning this since I noticed it while reviewing this PR.

lib/request-wrapper.ts
dpopp07: Yes, great call. I thought about that after I made the changes but I forgot to go back and implement it. I'm glad you pointed that out!

dpopp07:
Okay @padamstx, your comments have been addressed!

padamstx (approved):
LGTM

lib/helper.ts
```diff
@@ -340,3 +340,14 @@ export function constructServiceUrl(
...
+
+/**
+ * Returns true if and only if "mimeType" is a "JSON-like" mime type
+ * (e.g. "application/json; charset=utf-8").
+ * @param mimeType - the mimeType string
+ * @returns true if "mimeType" represents a JSON media type and false otherwise
+ */
+export function isJsonMimeType(mimeType: string) {
+  logger.debug(`Determining if the mime type '${mimeType}' specifies JSON content.`);
+  return !!mimeType && /^application\/json(;.*)?$/i.test(mimeType);
```
ricellis, line 352: small nit, I think it may also be valid for the `;` to be preceded by whitespace e.g.
The spec for [media type](https://www.rfc-editor.org/rfc/rfc9110.html#name-media-type) says:
>The type/subtype MAY be followed by semicolon-delimited parameters ([Section 5.6.6](https://www.rfc-editor.org/rfc/rfc9110.html#parameter)) in the form of name/value pairs.

In turn [5.6.6](https://www.rfc-editor.org/rfc/rfc9110.html#name-parameters) says:
>`parameters      = *( OWS ";" OWS [ parameter ] )`

where OWS is defined as optional white space by [5.6.3](https://www.rfc-editor.org/rfc/rfc9110.html#name-whitespace).

I can't say I've ever seen the first one preceded by whitespace, but it probably isn't worth tripping over.

lib/request-wrapper.ts
```diff
@@ -306,20 +307,17 @@ export class RequestWrapper {
...
-        errorBody = axiosError.data;
-        errorBody.warning = 'Body contains circular reference';
-        logger.error(`Failed to stringify axiosError: ${e}`);
+        logger.warn('Error field `result` contains circular reference(s)');
+        logger.debug(`Failed to stringify error response body: ${e}`);
       }
 
-      error.body = errorBody;
+      error.result = ensureJSONResponseBodyIsObject(axiosError);
+      error.body = error.result; // ** deprecated **
```
ricellis, line 320: Putting the `object` form of the response into the `result` field to match success cases opens an opportunity to reduce the impact of this otherwise breaking change when the type becomes an `object` instead of a `string`. I know, for instance, that in our SDK we have the `error.body` documented as "the error response body as text".

As such I wonder if it is worth leaving the deprecated `body` property as a `string` type i.e. `error.body = axiosError.data;` to preserve the behaviour for existing code rather than having 2 copies of the `object` form?

lib/helper.ts
dpopp07: Good catch! I'll make that adjustment

lib/request-wrapper.ts
dpopp07: Ah that is a great idea, thanks for the suggestion! I will do that 👍

lib/helper.ts
dpopp07: Regex updated and test added

lib/request-wrapper.ts
dpopp07: Reverted `error.body` to previous behavior and updated the tests

padamstx:
Latest changes look good (I already approved).

PR #239, opened 2023-03-01
fix: correctly handle empty response bodies

dpopp07:
There are cases, such as a HEAD request, where the response body will have a JSON content type but the body will be empty (i.e. an empty string). This is technically valid JSON but is not parsable into an object, so the code throws an exception. This commit adds specific handling for this scenario, as well as a couple of test cases


padamstx (approved):
LGTM

PR #241, opened 2023-05-03
fix(ContainerAuthenticator): add sa-token as default CR token filename

padamstx:
This PR modifies the ContainerAuthenticator so that it supports a second default CR token
filename.

##### Checklist

- [x] tests are included


pyrooka (approved):
LGTM!

dpopp07 (approved):
Looks good! And I confirmed this works with our internal cluster 👍

PR #247, opened 2023-07-03
feat(serializer): add serializeModel and createRequestAndDeserializeResponse

mojito317:
This PR introduces `createRequestAndDeserializeResponse` and `serializeModel` for serialization/deserialization feature.

##### Checklist

- [x] tests are included

### DIFF
--- a/lib/base-service.ts
+++ b/lib/base-service.ts
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2014, 2022.
+ * (C) Copyright IBM Corp. 2014, 2023.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lib/base-service.ts
```diff
@@ -255,6 +277,31 @@ export class BaseService {
...
 
+  /**
+   * Wrapper around `createRequest` that enforces arrived response to be deserialized.
+   * @param parameters - see `parameters` in `createRequest`
+   * @param deserializerFn - the deserializer function that is applied on the response object
+   * @param isMap - is `true` when the response object should be handled as a map
+   * @protected
+   * @returns a Promise
+   */
+  protected createRequestAndDeserializeResponse(
```
padamstx, line 288: The name of this function is `createRequestAndDeserializeResponse`, but yet it ends up calling a function named `serializeModel` after receiving the response body.  Shouldn't this be called `deserializeModel` instead?   Serialization would be like calling JSON.stringify() whereas deserialization would be like calling JSON.parse().

padamstx (commented):
I'm a little confused with the terminology...  The `createRequestAndDeserializeResponse` takes in a parameter named "deserializerFn", which it ends up passing to the `serializeModel` function.  Shouldn't `serializeModel` really be named `deserializeModel`???

lib/base-service.ts
mojito317: I agree, I renamed the function to the less confusing `convertModel` in [9c9e5cf](https://github.com/IBM/node-sdk-core/pull/247/commits/9c9e5cf4dfcf13dea654a284c62587042806f7ec).

padamstx (approved):
LGTM!

dpopp07 (approved):
Looks good! 👍

PR #256, opened 2023-10-30
fix(deps): resolve vulnerability in axios version

dpopp07:
This commit forces a release for the updated axios version that resolved a vulnerability. It also includes build changes to stop testing on Node 14, which had EOL 6 months ago, start testing on Node 20, and document that Node 14 is technically no longer guaranteed, even though it is supported.


padamstx (approved):
LGTM

PR #262, opened 2023-11-27
feat: improve HTTP redirects behavior

pyrooka:
This commit modifies the Node core so that it will include "safe" headers when performing a cross-site redirect where both the original and redirected hosts are within IBM's "cloud.ibm.com" domain.

##### Checklist

- [x] tests are included

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,6 +9,6 @@ Bug fixes and new features should include tests whenever possible.
 ##### Checklist
 <!-- Remove items that do not apply. For completed items, change [ ] to [x]. -->
 
-- [ ] `npm test` passes (tip: `npm run lint-fix` can correct most style issues)
+- [ ] `npm test` passes (tip: `npm run lint:fix` can correct most style issues)
 - [ ] tests are included
 - [ ] documentation is changed or added
--- a/lib/request-wrapper.ts
+++ b/lib/request-wrapper.ts
@@ -1,4 +1,4 @@
-/* eslint-disable class-methods-use-this */
+/* eslint-disable class-methods-use-this, no-restricted-syntax */
 
 /**
  * (C) Copyright IBM Corp. 2014, 2023.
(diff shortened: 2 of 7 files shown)


.github/pull_request_template.md
```diff
@@ -9,6 +9,6 @@ Bug fixes and new features should include tests whenever possible.
 ##### Checklist
 <!-- Remove items that do not apply. For completed items, change [ ] to [x]. -->
 
-- [ ] `npm test` passes (tip: `npm run lint-fix` can correct most style issues)
+- [ ] `npm test` passes (tip: `npm run lint:fix` can correct most style issues)
```
padamstx, line 12: good catch!

lib/request-wrapper.ts
```diff
@@ -580,3 +609,8 @@ function ensureJSONResponseBodyIsObject(response: any): any | string {
 
   return dataAsObject;
 }
+
+// Returns true iff safe headers should be copied to a redirected request.
+function shouldCopySafeHeadersOnRedirect(fromHost: string, toHost: string): boolean {
+  return fromHost.endsWith('.cloud.ibm.com') && toHost.endsWith('.cloud.ibm.com');
```
padamstx, line 615: This logic should work as long as we are also supporting the scenario where a redirection to the same host also results in the safe headers being copied, which I assume is the case since you're not completely overriding axios' redirection behavior (presumably) :)

test/unit/redirect.test.js
```diff
@@ -0,0 +1,211 @@
...
+  return { service, parameters };
+}
+
+describe('Node Core redirects', () => {
+  afterEach(() => {
+    nock.cleanAll();
+  });
+
+  it('should include safe headers within cloud.ibm.com domain', async () => {
+    const url1 = 'http://region1.cloud.ibm.com';
```
padamstx, line 57: let's use https in these URLs :)

test/unit/redirect.test.js
```diff
@@ -0,0 +1,211 @@
+/**
```
padamstx, line 1: need to add tests for redirects to the same host:
* both are region1.cloud.ibm.com
* both are region2.notcloud.ibm.com

padamstx (commented):
Looks good.
Only suggestion is to add a couple tests

test/unit/redirect.test.js
pyrooka: Agreed! I also change the protocols in my Python core PR, because I used HTTP there too.

lib/request-wrapper.ts
pyrooka: Those scenarios are covered by the unit tests now and yes, we are not overriding the whole redirection behavior, it's just a sort of post-processing in the chain, before sending the next, redirected request.

PR #267, opened 2024-02-01
fix(token-manager): handle request errors when refreshing tokens

dpopp07:
This commit introduces a Promise rejection handler to the `getToken` method where it was previously lacking. The refresh token kick-off did not catch any exceptions, resulting in the program crashing when a refresh error occurred. Now, the error will be caught and logged but the logic will continue as-is and the request will be retried later.

Note: I also discovered that one of our tests wasn't actually doing what it said it was doing 🙂 So I updated it to match its intentions and added a new test to verify the behavior that should have been captured by it.

##### Checklist

- [x] tests are included

### DIFF
--- a/test/unit/jwt-token-manager.test.js
+++ b/test/unit/jwt-token-manager.test.js
@@ -22,6 +22,7 @@ const decode = require('jsonwebtoken/decode');
 decode.mockImplementation(() => ({ exp: 100, iat: 100 }));
 
 const { JwtTokenManager } = require('../../dist/auth');
+const logger = require('../../dist/lib/logger').default;
 
 function getCurrentTime() {
   return Math.floor(Date.now() / 1000);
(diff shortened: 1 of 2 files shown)


test/unit/jwt-token-manager.test.js
```diff
@@ -141,14 +147,86 @@ describe('JWT Token Manager', () => {
...
+      expect(instance.accessToken).toBe(ACCESS_TOKEN);
+      expect(tokenNeedsRefreshSpy).not.toHaveBeenCalled();
+
+      saveTokenInfoSpy.mockRestore();
+      decode.mockRestore();
+      requestTokenSpy.mockRestore();
+      tokenNeedsRefreshSpy.mockRestore();
+    });
+
+    it('should refresh token if token is stored but needs refresh (close to expired)', async () => {
```
dpopp07, line 160: This now checks the "needs refresh but hasn't expired" path (the success version)

test/unit/jwt-token-manager.test.js
```diff
@@ -141,14 +147,86 @@ describe('JWT Token Manager', () => {
...
+      expect(instance.accessToken).toBe(ACCESS_TOKEN);
+      expect(decode).toHaveBeenCalled();
 
       saveTokenInfoSpy.mockRestore();
+      requestTokenSpy.mockRestore();
+      isTokenExpiredSpy.mockRestore();
       decode.mockRestore();
+    });
+
+    it('should return stored token without error if token refresh fails but should not save token info', async () => {
```
dpopp07, line 188: This checks the "needs refresh but hasn't expired" path (the failure version). This test would fail without my changes

padamstx (approved):
LGTM!

PR #268, opened 2024-02-27
fix: adjust IAM token expiration time

padamstx:
This commit changes the IAM, Container and VPC Instance authenticators slightly so that an IAM access token will be viewed as "expired" when the current time is within 10 seconds of the official expiration time. IOW, we'll expire the access token 10 secs earlier than the IAM server-computed expiration time.
We're doing this to avoid a scenario where
an IBM Cloud service receives a request along
with an "almost expired" access token and then uses that token to perform downstream requests in a
somewhat longer-running transaction and then the
access token expires while that transaction is
still active.

##### Checklist

- [x] tests are included

### DIFF
--- a/auth/token-managers/iam-request-based-token-manager.ts
+++ b/auth/token-managers/iam-request-based-token-manager.ts
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2019, 2023.
+ * (C) Copyright IBM Corp. 2019, 2024.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,13 +17,14 @@
 import extend from 'extend';
 import { OutgoingHttpHeaders } from 'http';
 import logger from '../../lib/logger';
-import { computeBasicAuthHeader, onlyOne, removeSuffix } from '../utils/helpers';
+import { computeBasicAuthHeader, getCurrentTime, onlyOne, removeSuffix } from '../utils/helpers';
 import { JwtTokenManager, JwtTokenManagerOptions } from './jwt-token-manager';
 
 const CLIENT_ID_SECRET_WARNING =
   'Warning: Client ID and Secret must BOTH be given, or the header will not be included.';
 const DEFAULT_IAM_URL = 'https://iam.cloud.ibm.com';
 const OPERATION_PATH = '/identity/token';
+const IAM_EXPIRATION_WINDOW = 10;
 
 /** Configuration options for IAM token retrieval. */
 export interface IamRequestOptions extends JwtTokenManagerOptions {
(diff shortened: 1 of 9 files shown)


auth/token-managers/iam-request-based-token-manager.ts
```diff
@@ -180,4 +181,24 @@ export class IamRequestBasedTokenManager extends JwtTokenManager {
...
+  /**
+   * Returns true iff the currently-cached IAM access token is expired.
+   * We'll consider an access token as expired when we reach its IAM server-reported
+   * expiration time minus our expiration window (10 secs).
+   * We do this to avoid using an access token that might expire in the middle of a long-running
+   * transaction within an IBM Cloud service.
+   *
+   * @returns true if the token has expired, false otherwise
+   */
+  protected isTokenExpired(): boolean {
```
padamstx, line 194: This is the key part of the fix. We'll override the inherited `isTokenExpired()` function so that it returns true as soon as we get within 10 seconds of
the IAM server-reported expiration time.

auth/token-managers/jwt-token-manager.ts
```diff
@@ -82,7 +82,7 @@ export class JwtTokenManager extends TokenManager {
 
     const decodedResponse = decode(this.accessToken);
     if (!decodedResponse) {
-      const err = 'Access token recieved is not a valid JWT';
+      const err = 'Access token received is not a valid JWT';
```
padamstx, line 85: I'm always trying to score points with @pyrooka :joy:

auth/token-managers/token-manager.ts
```diff
@@ -229,7 +229,7 @@ export class TokenManager {
    * Checks if currently-stored token should be refreshed
    * i.e. past the window to request a new token
    */
-  private tokenNeedsRefresh(): boolean {
+  protected tokenNeedsRefresh(): boolean {
```
padamstx, line 232: technically, I didn't need to override `tokenNeedsRefresh()` but it didn't seem right leaving it as the only private function :)

auth/token-managers/vpc-instance-token-manager.ts
```diff
@@ -174,4 +175,24 @@ export class VpcInstanceTokenManager extends JwtTokenManager {
...
+  protected isTokenExpired(): boolean {
+    const { expireTime } = this;
+
+    if (!expireTime) {
+      return true;
+    }
+
+    const currentTime = getCurrentTime();
+    return currentTime >= expireTime - IAM_EXPIRATION_WINDOW;
+  }
```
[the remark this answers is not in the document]
pyrooka: Fix me, but this token manager is inherited from the JWT token manager which doesn't have the correct method - no expiration window there.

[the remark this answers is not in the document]
dpopp07: Ah, right - I see. Thanks. My first thought is that perhaps we should be doing this for all JWT token managers. But I also realized Phil is only implementing this for certain authenticators so maybe this is fine as is.

pyrooka (approved):
LGTM!

auth/token-managers/jwt-token-manager.ts
pyrooka: Oh man, the amount of point I should give you, would cause a buffer overflow in any system out there... :)

auth/token-managers/vpc-instance-token-manager.ts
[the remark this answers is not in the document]
padamstx: Yes, I really only wanted to implement this for IAM-based authenticators, which are IAM, Container, and VPC Instance authenticators.

dpopp07 (approved):
Looks good! 👍

PR #270, opened 2024-04-10
build: bump/remove dependencies

padamstx:
This PR bumps a few of our dependencies and removes a couple of unneeded dependencies.

##### Checklist

- [ ] tests are included


package.json
```diff
@@ -71,16 +69,15 @@
...
     "@types/tough-cookie": "^4.0.0",
     "axios": "1.6.8",
-    "camelcase": "^5.3.1",
+    "camelcase": "^6.3.0",
     "debug": "^4.3.4",
-    "dotenv": "^6.2.0",
-    "expect": "^26.1.0",
+    "dotenv": "^16.4.5",
+    "expect": "^27.5.1",
     "extend": "^3.0.2",
```
dpopp07, line 76: One thing I thought of for this package is that, since it's such an old and stable version, we should hardcode the version. That way, if a compromised version is ever released as a fix (which I've seen happen for a big package before), we won't pull it in. If they were making regular updates, we'd want to include them, but since they aren't, we can require the stable version.

```suggestion
    "extend": "3.0.2",
```

dpopp07 (approved):
Looks good! 👍 

Left one suggestion about `extend` but it's small - approving to avoid re-review

PR #272, opened 2024-04-16
feat: send user-agent header with auth token requests

padamstx:
This commit updates our various request-based authenticators so that the User-Agent header is included with each outbound token request. The value of the User-Agent header will be of the form `ibm-node-sdk-core/<authenticator-type>-<core-version> <os-info>`.

Edit: I also updated `BaseService.createRequest()` so that it will add a default User-Agent header to the outbound request if one was not passed in with the request options.

##### Checklist

- [x] tests are included

### DIFF
--- a/auth/token-managers/cp4d-token-manager.ts
+++ b/auth/token-managers/cp4d-token-manager.ts
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2019, 2023.
+ * (C) Copyright IBM Corp. 2019, 2024.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 import extend from 'extend';
 import { validateInput } from '../utils/helpers';
+import { buildUserAgent } from '../../lib/build-user-agent';
 import { JwtTokenManager, JwtTokenManagerOptions } from './jwt-token-manager';
 
 /** Configuration options for CP4D token retrieval. */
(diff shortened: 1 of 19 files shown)


auth/token-managers/cp4d-token-manager.ts
```diff
@@ -16,6 +16,7 @@
 
 import extend from 'extend';
 import { validateInput } from '../utils/helpers';
+import { buildUserAgent } from '../../lib/privateHelper';
```
dpopp07, line 19: Convention is to use kebab-case file names.

Also, this is picky, but if we might add other private helpers to that file perhaps it should be plural? If not, maybe it should just be called `build-user-agent`.

```suggestion
import { buildUserAgent } from '../../lib/private-helpers';
```

lib/privateHelper.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
dpopp07, line 1: See previous note on filename

test/unit/token-manager-helper.js
```diff
@@ -0,0 +1,27 @@
+/**
```
dpopp07, line 1: The purpose of this might be more clear if it was in a `utils` file, or something, within the `test/unit` directory. It's perfectly fine here, too - just a thought.

dpopp07 (approved):
Looks good, aside from one request to change a file name, but I'll approve to avoid re-review

auth/token-managers/cp4d-token-manager.ts
padamstx: fixed

pyrooka (approved):
Looks good, as far as I can tell. :)

test/unit/token-manager-helper.js
padamstx: fixed

PR #275, opened 2024-05-30
fix(deps): remove `expect` as redundant development dependency

### LINKED ISSUE #174: SDK pulls in development level dependencies for production
This SDK is used by several other IBM cloud SDKs, in my case the @ibm-cloud/secrets-manager SDK. My application is a consumer of the @ibm-cloud/secrets-manager and I would like to package it (and its dependencies) as part of my app.

I notice that this causes a significant amount dependencies to be pulled in that are not required for production and that should also not appear in the dependencies of a production level application, e.g. the `expect` package. This package is meant to write unit tests and should only be a development level dependency. It alone adds around 200KB to the overall bundle size, because it in turn depends on `jest` packages which should not exist in a production environment either (see https://bundlephobia.com/package/expect@27.3.1)

The reason for this dependency is the fact that this core SDK exports utility functions for testing purposes as part of its regular index (https://github.com/IBM/node-sdk-core/blob/main/index.ts#L21)
Since this SDK is also only available as a commonjs package, there is also no way to use tree shaking to get rid of these undesirable and unused dependencies.

I suggest to move the test functions to a sub-package with its own `package.json` and its own dependencies. This way consuming SDKs can still make use of the exported test helpers but they would not interfer with the production usage of the package.

Besides the undesirable bundle size this also effects legal clearance when we at IBM depend on the SDK, because test level packages typically do not have to be cleared with the same diligence as production level packages. With the current design, we are forced to clear test packages.

### LINKED ISSUE #266: move some dependencies to devdependencies
Although this is not a big deal , there are a few packages currently reported as dependency, that rather should be dev dependencies as they are mostly used during build phase and not at run time

``` 
"dependencies": {
    "@types/debug": "^4.1.12",
    "@types/file-type": "~5.2.1",
    "@types/isstream": "^0.1.0",
    "@types/node": "~10.14.19",
    "@types/tough-cookie": "^4.0.0",
    "expect": "^26.1.0"
  },
```

thanks

dpopp07:
We previously exposed the `expect` package as a production-level dependency, because this package includes unit test utility functions for SDK libraries that depend on this package. However, `expect` is meant to be a development dependency and introduces some sizebloat and vulnerability attack surface to this package.

Fortunately, the `expect` package is intended to be used from `jest` and is not required to be installed on its own unless specifically needed. Since all SDK libraries depending on this package also use `jest`, it is redundant to include the dependency in this package. The test utilities provided will never be run in production and the SDKs will be able to run their tests using the already-installed `jest` package. For this reason, `expect` is removed in this commit.

Resolves #174 

Additionally, a couple of unnecessary dependencies are moved from the production list to the development list.

Resolves #266 

Note: I've tested this with our system tests and everything works fine without any changes to the SDKs.

### DIFF
--- a/lib/sdk-test-helpers.ts
+++ b/lib/sdk-test-helpers.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import expect from 'expect';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { expect } from '@jest/globals';
 
 /**
  * This module provides a set of helper methods used to reduce code duplication in the generated unit tests
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,6 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@types/debug": "^4.1.12",
-        "@types/file-type": "~5.2.1",
-        "@types/isstream": "^0.1.0",
         "@types/node": "~10.14.19",
         "@types/tough-cookie": "^4.0.0",
         "axios": "1.6.8",
@@ -36,6 +34,8 @@
         "@semantic-release/changelog": "6.0.3",
         "@semantic-release/git": "10.0.1",
         "@types/extend": "^3.0.4",
+        "@types/file-type": "~5.2.1",
+        "@types/isstream": "^0.1.0",
         "@types/mime-types": "^2.1.4",
         "@typescript-eslint/eslint-plugin": "^5.47.0",
         "@typescript-eslint/parser": "^5.47.0",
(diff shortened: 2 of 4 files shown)


lib/sdk-test-helpers.ts
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import expect from 'expect';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { expect } from '@jest/globals';
```
dpopp07, line 18: This import is changed from "expect" to "jest", since that is what will be available in the SDK projects by default.

padamstx (approved):
LGTM

PR #277, opened 2024-06-04
fix: revert "fix(deps): remove `expect` as redundant development dependency"

### LINKED ISSUE #276: Error: Cannot find module '@jest/globals'
Remember, an issue is not the place to ask questions. You can use
[Stack Overflow](http://stackoverflow.com/questions) for that, or
you may want to start a discussion on [dW Answers](https://developer.ibm.com/answers/questions/ask).

Before you open an issue, please check if a similar issue already exists or has been closed before.

Hi,
Since change #275 was merged, our downstream dependencies using this repo are failing with the following error:
```
Exception during run: Error: Cannot find module '@jest/globals'
code: 'MODULE_NOT_FOUND',
requireStack: [
  '/opt/ibm/app/node_modules/ibm-cloud-sdk-core/lib/sdk-test-helpers.js',
  '/opt/ibm/app/node_modules/ibm-cloud-sdk-core/index.js',
```
Assuming this would likely be breaking your internal building suite too.

I think the change here should be as simple as adding `@jest/globals` to your package.json and this should resolve things.
Thank you!

### When reporting a bug, please be sure to include the following:
- [ ] Steps to reproduce
- [x] Expected behavior - should install successfully
- [x] Actual behavior - fails to install
- [x] Node version - 18
- [ ] SDK version

### When you open an issue for a feature request, please add as much detail as possible:
- [x] A descriptive title starting with the service name
- [x] A description of the problem you're trying to solve
- [x] A suggested solution if possible

dpopp07:
This reverts commit b57b49108f17008fdf670ae2364edc908181773c.

Resolves #276

### DIFF
--- a/lib/sdk-test-helpers.ts
+++ b/lib/sdk-test-helpers.ts
@@ -14,8 +14,7 @@
  * limitations under the License.
  */
 
-// eslint-disable-next-line import/no-extraneous-dependencies
-import { expect } from '@jest/globals';
+import expect from 'expect';
 
 /**
  * This module provides a set of helper methods used to reduce code duplication in the generated unit tests
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "camelcase": "^6.3.0",
         "debug": "^4.3.4",
         "dotenv": "^16.4.5",
+        "expect": "^27.5.1",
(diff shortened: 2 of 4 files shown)


padamstx (approved):
LGTM

PR #278, opened 2024-06-06
v5 Release

### LINKED ISSUE #174: SDK pulls in development level dependencies for production

dpopp07:
This is the PR for releasing major version 5 of this package. There are two changes being made:
1) Updating the `engines` field to require Node 18 or later.
2) Removing the test utilities from the package and adding them via a separate package (this is the primary change).

### Notes on test utility changes

The SDK unit test helpers, which relied on test-scoped dependency "expect",
have been removed from the package. They will be exposed through a new NPM
package, separate from this one. The intention is to decouple testing
logic from production logic. This package, which is scoped for production,
will no longer contain the "expect" dependency, which has caused issues for
users. SDK projects will need to install the new package as a development
dependency and update their test code to use it instead. The functionality
will all remain the same.

This new package is still included in this project but it will not be built
and published as a part of public API. Instead, it will be isolated and
self-contained, with its own package.json, etc. For now, it will be published
manually - semantic release will only be used to automate publishing of the core.

This will be a breaking change. To continue using the test helpers, which SDK unit tests still rely on, users will need to install the new "@ibm-cloud/sdk-test-utilities" package as a development dependency.

Resolves #174 

fyi @ricellis @CarstenLeue


pyrooka (approved):
All changes look good to me!

PR #279, opened 2024-07-24
docs(README.md): update min version of node to be 18

padamstx:
This project's package.json has specified version 18 as the min node version for a while now.  This commit just updates the README.md to reflect that.


README.md
```diff
@@ -13,8 +13,8 @@ This project contains core functionality required by Node.js code generated by t
 ```
 
 ## Prerequisites
-- Node.js version 14 or newer
-- Note that v14 is technically *supported* but not officially *guaranteed* as we only test on v16 and higher
+- Node.js version 18 or newer
+- Note that v18 is technically *supported* but not officially *guaranteed* as we only test on v16 and higher
```
dpopp07, line 17: We can remove this line - our minimum version and our build have been aligned.

dpopp07 (changes requested):
Good catch! One small change to make though

dpopp07 (approved):
Looks good! 👍

README.md
padamstx: done

PR #280, opened 2024-07-29
build: add detect-secrets stage to build

padamstx:


.travis.yml
```diff
@@ -23,10 +20,25 @@ script:
...
+        - pip install --upgrade "git+https://github.com/ibm/detect-secrets.git@master#egg=detect-secrets"
+      script:
+        - detect-secrets scan --update .secrets.baseline
+        - detect-secrets -v audit --report --fail-on-unaudited --fail-on-live --fail-on-audited-real .secrets.baseline
+
+    - stage: Semantic-Release
+      language: node_js
+      node_js: 22
+      script:
+        - npm run semantic-release
```
dpopp07, line 44: We don't have a `semantic-release` NPM target in this repo, which is why we were using `npx` before. So we need to either update this script to use `npx` or add a script target in the `package.json` file like in [other cores](https://github.com/IBM/go-sdk-core/blob/main/package.json#L16). I'm happy either way

dpopp07 (changes requested):
Looks good but the semantic release step needs a fix

.travis.yml
pyrooka: I'm not the author, but nice catch. I don't think I would've been able to spot that... 😂

pyrooka (approved):
Besides that minor thing that Dustin found, it looks good.

.travis.yml
padamstx: I went ahead and added the script defn to package.json so that it's consistent with the other projects.

padamstx:
> Looks good but the semantic release step needs a fix

Fixed

dpopp07 (approved):
Looks good! 👍

PR #287, opened 2024-10-04
feat(IamAssumeAuthenticator): add new authentication type for iam assume

dpopp07:
This commit introduces the new IamAssumeAuthenticator which will fetch an IAM access token using the IAM getToken operation's "assume" grant type. The resulting access token allows the application to assume the identity of a trusted profile, similar to the "sudo" feature of Linux.

##### Checklist

- [x] tests are included

### DIFF
--- a/Authentication.md
+++ b/Authentication.md
@@ -2,7 +2,8 @@
 The node-sdk-core project supports the following types of authentication:
 - Basic Authentication
 - Bearer Token Authentication
-- Identity and Access Management (IAM) Authentication
+- Identity and Access Management (IAM) Authentication (grant type: apikey)
+- Identity and Access Management (IAM) Authentication (grant type: assume)
 - Container Authentication
 - VPC Instance Authentication
 - Cloud Pak for Data Authentication
@@ -16,7 +17,7 @@ which authentication types are supported for that service.
 
 The node-sdk-core allows an authenticator to be specified in one of two ways:
 1. programmatically - the SDK user invokes the appropriate function(s) to create an instance of the
-desired authenticator and then passes the authenticator instance when constructing an instance of the service.
+desired authenticator and then passes the authenticator instance when constructing an instance of the service client.
 2. configuration - the SDK user provides external configuration information (in the form of environment variables
 or a credentials file) to indicate the type of authenticator, along with the configuration of the necessary properties
 for that authenticator. The SDK user then invokes the configuration-based authenticator factory to construct an instance
@@ -28,7 +29,7 @@ which will include the following:
 - The properties associated with the authenticator
 - An example of how to construct the authenticator programmatically
 - An example of how to configure the authenticator through the use of external
-configuration information.  The configuration examples below will use
+configuration information. The configuration examples below will use
 environment variables, although the same properties could be specified in a
 credentials file instead.
 
@@ -143,16 +144,16 @@ const service = ExampleServiceV1.newInstance(options);
 
 Note that the use of external configuration is not as useful with the `BearerTokenAuthenticator` as it
 is for other authenticator types because bearer tokens typically need to be obtained and refreshed
-programmatically since they normally have a relatively short lifespan before they expire.  This
+programmatically since they normally have a relatively short lifespan before they expire. This
 authenticator type is intended for situations in which the application will be managing the bearer
 token itself in terms of initial acquisition and refreshing as needed.
 
 
-## Identity and Access Management (IAM) Authentication
-The `IamAuthenticator` will accept a user-supplied api key and will perform
+## Identity and Access Management (IAM) Authentication (grant type: apikey)
+The `IamAuthenticator` will accept a user-supplied apikey and will perform
 the necessary interactions with the IAM token service to obtain a suitable
-bearer token for the specified api key.  The authenticator will also obtain
-a new bearer token when the current token expires.  The bearer token is
+bearer token for the specified apikey. The authenticator will also obtain
+a new bearer token when the current token expires. The bearer token is
 then added to each outbound request in the `Authorization` header in the
 form:
 ```
@@ -161,7 +162,7 @@ form:
 
 ### Properties
 
-- apikey: (required) the IAM api key
+- apikey: (required) the IAM apikey to be used to obtain an IAM access token.
 
 - url: (optional) The base endpoint URL of the IAM token service.
 The default value of this property is the "prod" IAM token service endpoint
@@ -178,13 +179,13 @@ endpoint as well (`https://iam.test.cloud.ibm.com`).
 
 - clientId/clientSecret: (optional) The `clientId` and `clientSecret` fields are used to form a
 "basic auth" Authorization header for interactions with the IAM token server. If neither field
-is specified, then no Authorization header will be sent with token server requests.  These fields
+is specified, then no Authorization header will be sent with token server requests. These fields
 are optional, but must be specified together.
 
 - scope: (optional) the scope to be associated with the IAM access token.
 If not specified, then no scope wil be associated with the access token.
 
-- disableSslVerification: (optional) A flag that indicates whether verificaton of the server's SSL
+- disableSslVerification: (optional) A flag that indicates whether verification of the server's SSL
 certificate should be disabled or not. The default value is `false`.
 
 - headers: (optional) A set of key/value pairs that will be sent as HTTP headers in requests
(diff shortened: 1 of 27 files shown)


auth/authenticators/iam-assume-authenticator.ts
```diff
@@ -0,0 +1,100 @@
...
+ * token from a user-supplied apikey, then exchanges this initial IAM access token
+ * for another IAM access token that has "assumed the identity" of the specified
+ * trusted profile.
+ *
+ * The bearer token will be sent as an Authorization header in the form:
+ *
+ *      Authorization: Bearer \<bearer-token\>
+ */
+export class IamAssumeAuthenticator extends IamRequestBasedAuthenticator {
+  protected tokenManager: IamAssumeTokenManager;
```
dpopp07, line 57: The way that the inheritance is set up in this core made it fairly easy to implement everything, because the shared IAM-related logic is all handled automatically.

I decided not to include any state in this class, because it didn't seem necessary and I wanted to keep it lightweight. We can revisit that decision if need be. It can also be added later in a compatible way, if need be.

auth/token-managers/container-token-manager.ts
```diff
@@ -115,7 +115,7 @@ export class ContainerTokenManager extends IamRequestBasedTokenManager {
   /**
    * Request an IAM token using a compute resource token.
    */
-  protected async requestToken(): Promise<any> {
```
dpopp07, line 118: This must have been a typo or copy/paste error. It doesn't change behavior whatsoever, the function is still "asynchronous", but we don't use the `await` keyword in this function so it's unnecessary to include the `async` keyword (it's only purpose is to provide context when using `await`).

auth/token-managers/iam-assume-token-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+    // Create an instance of the IamTokenManager, which will be used to obtain
+    // an IAM access token for use in the "assume" token exchange. Most option
+    // names are shared between these token manager, and extraneous options will
+    // be ignored, so we can pass the options structure to that constructor as-is.
+    this.iamDelegate = new IamTokenManager(options);
+
+    // Set the grant type and user agent for this flavor of authentication.
+    this.formData.grant_type = 'urn:ibm:params:oauth:grant-type:assume';
+    this.userAgent = buildUserAgent('iam-assume-authenticator');
+  }
```
dpopp07, line 110: Like the authenticator, most of the logic is already taken care of elsewhere, so this was easy to implement. It's similar to the Container authenticator in that way.

Unlike the Container authenticator, I decided not to include any setters/getters for the member variables. We don't include them in the `IamAuthenticator` and I thought consistency with that class was most important. It also didn't seem necessary, but we can add them compatibly later if need be.

auth/utils/helpers.ts
```diff
@@ -91,20 +91,6 @@ export function getCurrentTime(): number {
...
- * or both are undefined.
- *
- * @param a - The first object
- * @param b - The second object
- * @returns true if and only if exactly one of a or b is defined
- */
-export function onlyOne(a: any, b: any): boolean {
-  return Boolean((a && !b) || (b && !a));
-}
-
```
dpopp07, line 107: I just moved this to be further down in the file, next to the other related functions.

auth/utils/helpers.ts
```diff
@@ -122,25 +108,49 @@ export function removeSuffix(str: string, suffix: string): string {
...
- * @param a - the first object
- * @param b - the second object
- * @returns true if a or b is defined; false if both are undefined
+ * @param args - The spread of arguments to check
+ * @returns true if and only if exactly one argument is defined
  */
-export function atLeastOne(a: any, b: any): boolean {
-  return Boolean(a || b);
+export function onlyOne(...args: any): boolean {
+  return countDefinedArgs(args) === 1;
```
dpopp07, line 119: I changed all of these functions to accept spread arguments instead of binary options.

The check needed for this authenticator (exactly one out of three) prompted me to think these functions would be more valuable down the road if they weren't limited to looking at only two options. So, now they can take any number of options and still perform the same checks. They are perfectly compatible with their old implementations, though they've technically "changed" in the API.

etc/ibm-cloud-sdk-core.api.md
```diff
@@ -13,10 +13,10 @@ import { OutgoingHttpHeaders } from 'http';
 import { Stream } from 'stream';
 
 // @public
-export function atLeastOne(a: any, b: any): boolean;
+export function atLeastOne(...args: any): boolean;
```
dpopp07, line 16: Here we see the API change but note that they are compatible: `...args: any` can very well equal `a: any, b: any`.

auth/authenticators/iam-assume-authenticator.ts
```diff
@@ -0,0 +1,100 @@
...
+/** Configuration options for IAM Assume authentication. */
+export interface Options extends IamRequestOptions {
+  /** The IAM api key */
+  apikey: string;
+
+  /**
+   * Specify exactly one of [iamProfileId, iamProfileCrn, or iamProfileName] to
+   * identify the trusted profile whose identity should be used. If iamProfileId
+   * or iamProfileCrn is used, the trusted profile must exist in the same account.
+   * If and only if IAMProfileName is used, then iamAccountId must also be
```
pyrooka, line 30:
```suggestion
   * If and only if iamProfileName is used, then iamAccountId must also be
```
Just to be consistent with the rest. :)

Authentication.md
```diff
@@ -28,7 +29,7 @@ which will include the following:
 - The properties associated with the authenticator
 - An example of how to construct the authenticator programmatically
 - An example of how to configure the authenticator through the use of external
-configuration information.  The configuration examples below will use
+configuration information. The configuration examples below will use
```
pyrooka, line 32: 🥇

Authentication.md
```diff
@@ -228,6 +229,130 @@ const service = ExampleServiceV1.newInstance(options);
...
+
+- clientId/clientSecret: (optional) The `clientId` and `clientSecret` fields are used to form a 
+"basic auth" Authorization header for interactions with the IAM token server when fetching the
+initial IAM access token. These fields are optional, but must be specified together.
+
+- scope: (optional) the scope to be used when obtaining the initial IAM access token.
+If not specified, then no scope will be associated with the access token.
+
+- disableSSLVerification: (optional) A flag that indicates whether verification of the server's SSL 
+certificate should be disabled or not. The default value is `false`.
```
pyrooka, line 291:
```suggestion
- disableSslVerification: (optional) A flag that indicates whether verification of the server's SSL 
certificate should be disabled or not. The default value is `false`.
```

Authentication.md
```diff
@@ -228,6 +229,130 @@ const service = ExampleServiceV1.newInstance(options);
...
+
+- headers: (optional) A set of key/value pairs that will be sent as HTTP headers in requests
+made to the IAM token service.
+
+### Usage Notes
+- The IamAssumeAuthenticator is used to obtain an access token (a bearer token) from the IAM token service
+that allows an application to "assume" the identity of a trusted profile.
+
+- The authenticator first uses the apikey, url, clientId/clientSecret, scope, disableSSLVerification, and headers
```
pyrooka, line 300:
```suggestion
- The authenticator first uses the apikey, url, clientId/clientSecret, scope, disableSslVerification, and headers
```

pyrooka (approved):
Looks good, only a few minor things to fix before merging.

test/unit/iam-assume-authenticator.test.js
```diff
@@ -0,0 +1,138 @@
...
+const mockedTokenManager = new IamAssumeTokenManager({
+  apikey: 'some-key',
+  iamProfileId: 'some-id',
+});
+
+const getTokenSpy = jest
+  .spyOn(mockedTokenManager, 'getToken')
+  .mockImplementation(() => Promise.resolve(fakeToken));
+
+describe('Container Authenticator', () => {
```
pyrooka, line 31:
```suggestion
describe('Container Authenticator', () => {
```
```suggestion
describe('IAM Assume Authenticator', () => {
```
Copy-paste?

test/unit/iam-assume-authenticator.test.js
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  // "End to end" style test, to make sure this authenticator ingregates properly with parent classes.
+  it('should update the options and resolve with `null` when `authenticate` is called', async () => {
+    const authenticator = new IamAssumeAuthenticator({
+      apikey: config.apikey,
+      iamProfileId: config.iamProfileId,
+    });
+
+    // Vverride the created token manager with the mocked one.
```
pyrooka, line 114:
```suggestion
    // Override the created token manager with the mocked one.
```
Just found another typo. 👮

test/integration/iam-assume-authenticator.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+// Note: Only the unit tests are run by default.
+//
+// In order to test with a live IAM server, create file "iamassume.env" in the project root.
+// It should look something like this:
+//
+//	ASSUMETEST_AUTH_URL=<url>   e.g. https://iam.cloud.ibm.com
+//	ASSUMETEST_AUTH_TYPE=iamAssume
+//	ASSUMETEST_APIKEY=<apikey>
+//  ASSUMETEST_IAM_PROFILE_ID=<profile-id>
```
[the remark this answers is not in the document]
dpopp07: Actually, it should be spaces everywhere. I will fix that locally, thanks!

test/unit/iam-assume-authenticator.test.js
dpopp07: You have 🦅 👀 sir! 🙂

test/unit/iam-assume-authenticator.test.js
```diff
@@ -0,0 +1,138 @@
...
+    expect(() => {
+      const unused = new IamAssumeAuthenticator({
+        apikey: config.apikey,
+        iamProfileId: 'some-id',
+        iamAccountId: 'some-account',
+      });
+    }).toThrow('`iamProfileName` and `iamAccountId` must be provided together, or not at all');
+  });
+
+  // "End to end" style test, to make sure this authenticator ingregates properly with parent classes.
```
padamstx, line 107:
```suggestion
  // "End to end" style test, to make sure this authenticator integrates properly with parent classes.
```
Sorry, but with @pyrooka on the review list, I need to demonstrate that I'm paying attention too :joy:

test/unit/iam-assume-authenticator.test.js
```diff
@@ -0,0 +1,138 @@
...
+
+    expect(result).toBeUndefined();
+    expect(options.headers.Authorization).toBe(`Bearer ${fakeToken}`);
+    expect(getTokenSpy).toHaveBeenCalled();
+
+    // Verify that the original options are kept intact.
+    expect(options.headers['X-Some-Header']).toBe('user-supplied header');
+  });
+
+  it('should return the refresh token stored in the token manager', () => {
```
padamstx, line 128: We're not actually supporting a refresh token returned by the getToken "assume" operation.  It's part of the response, but the new authenticator doesn't (or probably shouldn't) expose the refresh token.

padamstx (commented):
Looks good.  I found one small typo, and also added a comment about the refreshToken, which we're not intending to support...   or at least we do not yet support it in the Go and Java implementations.

test/unit/iam-assume-authenticator.test.js
dpopp07: Hard to see what he misses 😂 good catch!

auth/token-managers/iam-request-based-token-manager.ts
```diff
@@ -41,11 +41,11 @@ export interface IamRequestOptions extends JwtTokenManagerOptions {
...
  */
 export class IamRequestBasedTokenManager extends JwtTokenManager {
-  private clientId: string;
+  protected clientId: string;
 
-  private clientSecret: string;
+  protected clientSecret: string;
 
-  private scope: string;
+  protected scope: string;
```
dpopp07, line 48: I adjusted the visibility of these so that I can override them in the IAM Assume token manager (otherwise, they would be added to the request if set).

package.json
```diff
@@ -98,7 +98,7 @@
     "eslint:fix": "eslint . --fix",
     "eslint:check": "eslint . --cache",
     "lint": "npm run eslint:check",
-    "lint:fix": "npm run eslint:fix",
+    "fix": "npm run eslint:fix",
```
dpopp07, line 101: @padamstx - this change is for you 😂 I remember you recently pointing out our inconsistency between this project and the validator... now, resolved!

test/unit/iam-assume-authenticator.test.js
dpopp07: As discussed on Slack, the refresh token will still be "available" but it will not be stored when we process the response. I've removed this test.

padamstx (approved):
LGTM

test/unit/iam-assume-authenticator.test.js
pyrooka: Wow, how do you that?
(I had a vision correction surgery years ago and the name of the clinic where it was done is "Sasszemklinika" which means "EagleEyeClinic". 😂)

test/unit/iam-assume-authenticator.test.js
pyrooka: > Sorry, but with @pyrooka on the review list, I need to demonstrate that I'm paying attention too 😂

I left some for you on purpose, hahaha!

pyrooka (approved):
Still looks good! :)

test/unit/iam-assume-authenticator.test.js
dpopp07: That's actually crazy 😂 I had no idea!

padamstx (approved):
Latest changes look good.

PR #288, opened 2024-11-05
feat(sse): add support for server side events

apaparazzi0329:
This commit adds support for http server side event streaming by setting the axios responseType to 'stream' when the Accept header is 'text/event-stream'


dpopp07:
Thanks for the PR @apaparazzi0329 !

As discussed offline, this may be better suited to a change in the SDK itself. I'm going to close this for now.

PR #289, opened 2025-01-10
docs: update api.md file to reflect latest code

padamstx:


etc/ibm-cloud-sdk-core.api.md
```diff
@@ -440,21 +447,12 @@ export type TokenManagerOptions = {
     [propName: string]: any;
 };
 
+// Warning: (ae-forgotten-export) The symbol "TokenRequestBasedAuthenticatorImmutable" needs to be exported by the entry point index.d.ts
```
padamstx, line 450: Sort of worried about this.  Sounds ominous :joy:

padamstx: After poking around a bit, it looks like this warning is due to the fact that auth/authenticators/index.ts does not explicitly export `TokenRequestBasedAuthenticatorImmutable`.
Same warning for `IamRequestBasedAuthenticatorImmutable` also.

I think we need to export those two classes since they are referenced by other exported classes, right?

padamstx: I added the two immutable classes to the index.ts file and now everything seems to look better.

dpopp07: > I think we need to export those two classes since they are referenced by other exported classes, right?

IIUC just because we use them in classes we export doesn't mean they need to be exported. I can't remember if we intended these files to be used directly by SDK users or not. Did we? If not, we should be okay to continue leaving them un-exported.

padamstx: To close the loop on this, after discussing with @dpopp07, I removed the two exports for the immutable classes.  It turns out we have other instances of these warnings already, and that doesn't seem to cause any problems.

dpopp07 (approved):
Looks good! 👍

PR #295, opened 2025-02-17
fix(build): fix .releaserc format

padamstx:
Fixes the format of the .releaserc file to comply with the semrel docs.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,26 +1,19 @@
 {
+  "debug": true,
   "branches": [
     "main"
   ],
-  "verifyConditions": [
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
     "@semantic-release/changelog",
-    "@semantic-release/npm",
-    "@semantic-release/git"
-  ],
-  "prepare": [
-    "@semantic-release/changelog",
-    "@semantic-release/npm",
-    "@semantic-release/git"
-  ],
-  "publish": [
     [
       "@semantic-release/npm",
       {
         "pkgRoot": "dist"
       }
     ],
-    {
-      "path": "@semantic-release/github"
-    }
+    "@semantic-release/git",
+    "@semantic-release/github"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "sideEffects": false,
   "repository": {
     "type": "git",
-    "url": "https://github.com/IBM/node-sdk-core.git"
+    "url": "git+https://github.com/IBM/node-sdk-core.git"
   },
   "keywords": [
     "ibm",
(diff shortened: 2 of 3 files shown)


.releaserc
```diff
@@ -1,26 +1,19 @@
 {
+  "debug": true,
```
padamstx, line 2: I like to always enable debug with semantic-release because it doesn't add an order of magnitude to the volume of output and it's handy to have some of that extra info already displayed in the build logs if something goes wrong :)

.releaserc
```diff
@@ -1,26 +1,19 @@
...
   "branches": [
     "main"
   ],
-  "verifyConditions": [
+  "plugins": [
+    "@semantic-release/commit-analyzer",
```
padamstx, line 7: These first two are automatically included in "plugins" if no other configured plugin implements their primary steps (analyzeCommits and generateNotes, respectively), but I added them explicitly just to be consistent with .releaserc files in the other languages.

package.json
```diff
@@ -7,7 +7,7 @@
   "sideEffects": false,
   "repository": {
     "type": "git",
-    "url": "https://github.com/IBM/node-sdk-core.git"
+    "url": "git+https://github.com/IBM/node-sdk-core.git"
```
padamstx, line 10: This change was actually suggested in the warning messages displayed by `npm publish` in the build log of a recent main branch build :joy:
image
Link: https://github.com/IBM/node-sdk-core/actions/runs/13032430882/job/36354639235
So... I ran `npm pkg fix` and it made the above change.

package.json
```diff
@@ -61,7 +61,7 @@
     "npm-run-all": "4.1.5",
     "package-json-reducer": "1.0.18",
     "prettier": "~2.3.0",
-    "semantic-release": "^23.0.7",
+    "semantic-release": "24.2.3",
```
padamstx, line 64: Might as well live on the bleeding edge :joy:

dpopp07: Is there a reason you pinned the dependency, rather than using `^`?

package.json
padamstx: It was just a coincidence that I was paying particularly close attention to the logs since I wanted to study the semantic-release messages :joy:

package.json
padamstx: Not a good enough reason so that I could defend it :joy:   I'll add it back.

padamstx: updated it

dpopp07 (approved):
Looks good! 👍 

Though I would use a "build" commit since we aren't actually releasing any changes to the package. (I know that leaves us on the edge of our seat until the next actual change rolls through 🙂 so up to you).

padamstx:
> Though I would use a "build" commit since we aren't actually releasing any changes to the package. (I know that leaves us on the edge of our seat until the next actual change rolls through 🙂 so up to you).

I thought about that, but decided on a fix commit to force a new patch level, because I want to get these changes "tested" so I can then go ahead and make similar changes in some of the other node projects (template, node sdk, etc.).

PR #298, opened 2025-03-05
fix: update node types to align with required node versions

### LINKED ISSUE #297: @types/node should be in devDependencies instead of a dependencies
In package.json there is:

  "dependencies": 
    "@types/node": "~10.14.19",

This causing a direct dependency of a very old @types/node package while node 18 is the min requirement. It prevents consumers from being able to  cleanly use functions like rmSync() which was added since node 14.
Can it be moved to devDependencies or at least update to a more recent version?

Last year there was a similar [commit](https://github.com/IBM/node-sdk-core/commit/14cfce9e390b316b07216aa7b03cdc4eb91d6447) that moved some packages to devDependencies and explaining the reason for the move.

dpopp07:
Resolves #297

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
   },
   "dependencies": {
     "@types/debug": "^4.1.12",
-    "@types/node": "~10.14.19",
+    "@types/node": "^22.13.9",
     "@types/tough-cookie": "^4.0.0",
     "axios": "1.7.9",
     "camelcase": "^6.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
   },
   "dependencies": {
     "@types/debug": "^4.1.12",
-    "@types/node": "~10.14.19",
+    "@types/node": "^22.13.9",
```
padamstx, line 69: wow, that's a sizeable version bump :joy:

padamstx (approved):
LGTM

package.json
dpopp07: It sure is 😎 We've neglected updating that version

PR #299, opened 2025-03-06
feat: add support for accepting axios options in sendRequest

dpopp07:
This adds logic to extract the Axios options from the request parameters and pass them to the final request. It also exposes a type for generated SDKs to utilize, so they don't need to install Axios just for that type.

##### Checklist

- [x] tests are included


pyrooka (approved):
Looks good. Although the `AbortSignal_2` naming in the API Report looks a bit odd to me, I'm pretty sure it's correct and it's just my lack of TS knowledge that causes the confusion... :)

dpopp07:
> the AbortSignal_2 naming in the API Report looks a bit odd to me

Yes, that's a bit odd - it's a part of the automated docs generation. It's not exposed to the user. I guess the docs generated needs to have an intermediate step when renaming something 🤷‍♂️

padamstx (approved):
LGTM

PR #303, opened 2025-05-29
feat(auth): add support for MCSP V2 authentication

padamstx:
This commit adds the McspV2Authenticator implementation. This authenticator will invoke the MCSP V2 POST
/api/2.0/{scopeCollectionType}/{scopeId}/apikeys/token operation to obtain an access token for an apikey.

### DIFF
--- a/auth/token-managers/token-manager.ts
+++ b/auth/token-managers/token-manager.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars, class-methods-use-this */
 
 /**
- * (C) Copyright IBM Corp. 2020, 2024.
+ * (C) Copyright IBM Corp. 2020, 2025.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


auth/token-managers/token-manager.ts
```diff
@@ -199,7 +199,7 @@ export class TokenManager {
    * @returns Promise
    */
   protected requestToken(): Promise<any> {
-    const errMsg = '`requestToken` MUST be overridden by a subclass of TokenManagerV1.';
+    const errMsg = '`requestToken` MUST be overridden by a subclass of TokenManager.';
```
padamstx, line 202: This is how you know I was paying attention :joy:

lib/request-wrapper.ts
```diff
@@ -156,11 +156,15 @@ export class RequestWrapper {
    * @returns the string representation of the request
    */
   private formatAxiosRequest(request: InternalAxiosRequestConfig): string {
-    const { method, url, data, headers } = request;
-
+    const { method, url, data, headers, params } = request;
```
padamstx, line 159: This change just adds the query parameters to the log output, which came in handy while verifying that the authenticator is working!

PR #321, opened 2026-04-02
fix(deps): lock axios to fix CVE

diatrcz:
Newest version of axios (1.14.1) has a severe vulnerability so we need to lock the axios version 1.14.0.


diatrcz:
@pyrooka since this is a severe vulnerability and it can't wait until Tuesday I'm merging this without a review.

pyrooka (approved):
@diatrcz Ok, go ahead! ;)

PR #323, opened 2026-04-10
fix(deps): Bump axios to avoid CVEs

diatrcz:
Bumps (and locks) axios to 1.15.0 to avoid https://github.com/axios/axios/security/advisories/GHSA-3p68-rc4w-qgx5.


pyrooka (approved):
LGTM!

PR #325, opened 2026-04-14
test: add test case for nested objects

diatrcz:
Adds test case to verify that nested configuration options didn't get overwritten with the changes in the code.
Since we are planning on cutting down the npm package uses in the project, this test can help further verify that the behaviour of the SDK Core haven't changed even in edge cases.


pyrooka (approved):
Looks pretty good!

PR #342, opened 2026-04-30
fix(deps): bump and lock dependencies to fix vulnerabilties

diatrcz:
Due to the recent CVEs this PR locks the dependency version to prevent security breaches like the recent axios one.

It also bumps most dependencies that cause vulnerabilities shown in `npm audit` and `dependabot`.


package.json
```diff
@@ -38,57 +38,61 @@
...
+    "eslint": "7.26.0",
+    "eslint-config-airbnb-base": "14.2.1",
+    "eslint-config-airbnb-typescript": "12.3.1",
+    "eslint-config-prettier": "8.3.0",
+    "eslint-plugin-import": "2.23.2",
+    "eslint-plugin-jest": "24.3.6",
+    "eslint-plugin-jsdoc": "34.6.3",
+    "eslint-plugin-node": "9.0.0",
+    "eslint-plugin-prettier": "3.0.1",
     "jest": "^29.7.0",
```
pyrooka: Is there a reason you left `jest` out from the version pinning?

pyrooka (approved):
Left one questions, but it looks good overall!

package.json
diatrcz: Nope, I fixed it before merging the PR.

PR #352, opened 2026-05-13
fix: add cloning to deepmerge

diatrcz:
Adds deep cloning to the `deepMerge` function, this way arrays should be handeled correctly.

##### Checklist

- [x] tests are included


pyrooka (approved):
Looks good!

PR #353, opened 2026-05-14
fix(deps): revert extend removal

diatrcz:
Reverts the changes that removed the `extend` dependency from the project.


pyrooka (approved):
Looks good! 🤞

PR #358, opened 2026-05-21
build: bump axios to 1.16.1

Andris28:
In this commit we bump axios version
to 1.16.1 as it contains security fixes.


diatrcz (approved):
lgtm!

PR #359, opened 2026-06-04
feat(auth): add support for new service version in VPC Instance Auth

diatrcz:
This PR adds support for the new VPC Instance Metadata Service version (2025-08-26) with the following changes:

- Added configurable `ServiceVersion` field (default: "2022-03-01") that switches to new API paths (`/identity/v1/token`, `/identity/v1/iam_tokens`) when set to "2025-08-26"
- Added `TokenLifetime` field (default: 300s) for customizable token expiration
- Configuration via builder methods (`SetServiceVersion()`, `SetTokenLifetime()`) or `VPC_IMS_VERSION` environment variable
- **Full backward compatibility with older service versions**

##### Checklist

- [x] tests are included


Authentication.md
```diff
@@ -526,13 +533,16 @@ External configuration:
...
+export EXAMPLE_SERVICE_VERSION=2025-08-26
 ```
 Application code:
 ```js
 const ExampleServiceV1 = require('<sdk-package-name>/example-service/v1');
 
 const options = {
   serviceName: 'example_service',
+  serviceVersion: '2025-08-26',
+  tokenLifetime: 600,
```
pyrooka: Would you mind adding new examples that demonstrate the authenticator initiation with the new version/token lifetime?

auth/token-managers/vpc-instance-token-manager.ts
```diff
@@ -108,6 +130,33 @@ export class VpcInstanceTokenManager extends JwtTokenManager {
...
+    }
+    this.serviceVersion = serviceVersion;
+  }
+
+  public setTokenLifetime(tokenLifetime: number): void {
+    this.tokenLifetime = tokenLifetime;
+  }
+
+  protected getAccessTokenPath(): string {
+    if (this.serviceVersion === '2025-08-26') {
```
pyrooka: Similarly to the Java SDK Core PR, I think it would be worth putting this value into a const.

auth/token-managers/vpc-instance-token-manager.ts
```diff
@@ -108,6 +130,33 @@ export class VpcInstanceTokenManager extends JwtTokenManager {
...
+    if (!metadataServiceSupportedVersions.includes(serviceVersion)) {
+      throw new Error(
+        `Invalid serviceVersion. Must be one of: ${metadataServiceSupportedVersions.join(', ')}`
+      );
+    }
+    this.serviceVersion = serviceVersion;
+  }
+
+  public setTokenLifetime(tokenLifetime: number): void {
+    this.tokenLifetime = tokenLifetime;
```
pyrooka, line 155: What do you think about adding a validation for the `tokenLifetime` - like you did with the `serviceVersion`? Just to make sure it's not possible to set invalid values.

auth/authenticators/vpc-instance-authenticator.ts
```diff
@@ -24,6 +24,10 @@ export interface Options extends BaseOptions {
   iamProfileCrn?: string;
   /** The ID of the linked trusted IAM profile to be used when obtaining the IAM access token */
   iamProfileId?: string;
+
+  serviceVersion: string;
+
+  tokenLifetime: number;
```
pyrooka: These properties should be marked as optional. Also, could you add some comments to them?

```suggestion
  /** ...
  serviceVersion?: string;
  /** ...
  tokenLifetime?: number;
```

auth/token-managers/vpc-instance-token-manager.ts
```diff
@@ -22,13 +22,23 @@ import { JwtTokenManager, JwtTokenManagerOptions } from './jwt-token-manager';
...
 /** Configuration options for VPC token retrieval. */
 interface Options extends JwtTokenManagerOptions {
   /** The CRN of the linked trusted IAM profile to be used as the identity of the compute resource */
   iamProfileCrn?: string;
   /** The ID of the linked trusted IAM profile to be used when obtaining the IAM access token */
   iamProfileId?: string;
+
+  serviceVersion: string;
+
+  tokenLifetime: number;
```
pyrooka: These properties should be marked as optional. Also, could you add some comments to them?

```suggestion
  /** ...
  serviceVersion?: string;
  /** ...
  tokenLifetime?: number;
```

auth/token-managers/vpc-instance-token-manager.ts
```diff
@@ -81,6 +95,14 @@ export class VpcInstanceTokenManager extends JwtTokenManager {
     }
 
     this.url = options.url || DEFAULT_IMS_ENDPOINT;
+    this.serviceVersion = options.serviceVersion || METADATA_SERVICE_VERSION;
```
pyrooka: We should validate the service version in the constructor too. If the user doesn't use the setter method, the service version is not checked in the current implementation, right?

pyrooka (changes requested):
Overall it's pretty good, just a few comments to address before merging.

pyrooka (approved):
LGTM!

